Issue #13, opened 2022-12-22
Standardize and document Component export style


garyray-k:
Another comment for code standardization to be documented/adhered to.
Using `export default (props) =>` as opposed to `const ComponentName = (props) => ... export default ComponentName`

_Originally posted by @garyray-k in https://github.com/satsigner/satsigner/pull/7#discussion_r1054385766_

tmakerman:
I'd suggest we use 
`export default class Foobar`

and named functions
`export default function Baz`

I intentionally didn't write argument list or generic types since I don't have opinion on that yet :)

https://stackoverflow.com/a/60106119
https://github.com/airbnb/javascript/tree/master/react#declaration

tmakerman:
Also, suggest we pick a popular style guide as basis for project conventions and add addendum as necessary.  Looks like these ones are popular

https://github.com/piotrwitek/react-redux-typescript-guide
https://github.com/airbnb/javascript/tree/master/react

garyray-k:
> Also, suggest we pick a popular style guide as basis for project conventions and add addendum as necessary. Looks like these ones are popular
> 
> https://github.com/piotrwitek/react-redux-typescript-guide https://github.com/airbnb/javascript/tree/master/react

Good idea! I'm ambivalent about the actual style but very opinionated that we should have one for consistency. 😄

Psycarlo:
This is being done on the migration branch.

PR #16, opened 2022-12-23
Add view wrapper around Navigation Container

### LINKED ISSUE #15: Add View wrapper around Navigation component
https://stackoverflow.com/questions/59900898/white-background-flashing-when-switching-screens-react-navigation-v5

Look like we can add a `View` wrapper around Navigation to resolve this. Would you kindly address this in the PR?

- [x] Update with wrapper around `Navigation` component

_Originally posted by @garyray-k in https://github.com/satsigner/satsigner/pull/7#discussion_r1055871098_

tmakerman:
Previously was using a tranparent modal setting to get rid of white flash seen when navigating between screens.  Wrapping in View with same background color is better solution.

Closes #15

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,16 @@
 import React from 'react';
+import {
+  View,
+  StyleSheet
+} from 'react-native';
 
 import {NavigationContainer} from '@react-navigation/native';
 import {createNativeStackNavigator} from '@react-navigation/native-stack';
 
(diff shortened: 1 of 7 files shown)


tmakerman:
commit [7b1178c](https://github.com/satsigner/satsigner/pull/16/commits/7b1178c7eff24b83313eeab59ece0188becb7d53) also in this PR relates to #13

PR #18, opened 2023-07-11
Fix markdown issues

dergigi:
Replace all instances of `] (` with `](`. This removes the space so the markdown links render correctly.


tmakerman:
Thanks!

PR #32, opened 2024-04-04
[REFACTOR]  Functional Components (Home | AccountsComponent)

Fahad-Mahmood:
1- Files Converted to Functional Components
- App.tsx
- HomeScreen.tsx
- AccountComponents.tsx

2- Svg files import typescript warning fix (type declaration added)

3- AppText component missing prop types (style prop)

4- Prettier bracket spacing added (most of the files were already using this).


Psycarlo:
LGTM

Psycarlo (approved):

tmakerman:
> * These styles are used by two files (AccountList, and AccountComponent). I'll see if I can separate them (in the next PR maybe when I convert AccountList component).

That would be great.

tmakerman:
On the component called `AccountComponents`...  For components so far I have been adding `Screen` to end of component name but otherwise leaving `Component` out of component name.  Maybe `AccountComponents` could be renamed to `Accounts`?

tmakerman:
@Fahad-Mahmood I'm done reviewing.  This looks great.  Just the comment above on AccountComponents naming.

tmakerman (changes requested):
@Fahad-Mahmood I'm done reviewing.  This looks great.  Just the comment above on AccountComponents naming.

Fahad-Mahmood:
nice idea.

tmakerman (changes requested):
Sorry to be a pain :) but I think plural "Accounts" rather than "Account" as the component represents multiple accounts.

Fahad-Mahmood:
oh, that's correct! we receive an array prop for accounts

Fahad-Mahmood:
But I think we should create a separate directory for screens in the future.

tmakerman:
> But I think we should create a separate directory for screens in the future.

I think that makes sense, especially when multiple files are needed for a screen.  I think it will be a good pattern.

Fahad-Mahmood:
@Psycarlo @tmakerman Thanks for the review ❤️

PR #37, opened 2024-04-20
Selected UTXOs Header

tmakerman:
I'd ask for reviews on this branch for now to focus on the selected UTXOs header `src/components/accounts/SelectedUtxosHeader.tsx` - this to be shared by both the UTXOs listing screen and UTXO bubble view screen

Subsequent reviews on this code can focus more on the UTXOs listing screen, but its currently just a thrown together UI to test out the header component.

For now I put this header component under src/components/accounts, but later as we reorganize the codebase into the new `src/screens` structure it will move.


src/screens/AccountUtxoList/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+interface Props {
+  navigation: NavigationProp<any>;
+}
+
+export default function AccountUtxoListScreen({
+  navigation
+}: Props) {
+  const accountsContext = useAccountsContext();
+  const { currentAccount } = accountsContext;
```
Fahad-Mahmood, line 23: you can destructure objects in a single line:

```
const { currentAccount, utxos } = useAccountsContext();
```

src/components/accounts/TransactionBuilderProvider.tsx
```diff
@@ -1,27 +1,40 @@
...
+import { TransactionBuilderContext } from "./TransactionBuilderContext";
+
 export const TransactionBuilderProvider = ({ children }) => {
   
-  const [inputs, setInputs] = React.useState<Map<string, Utxo>>(new Map<string, Utxo>());
+  // using object indirection so assignment creates new container object and forces re-render
+  const [inputsMap, setInputsMap] = useState<{ map: Map<string, Utxo> }>({ map: new Map<string, Utxo>() });
 
-  const getOutpoint = (utxo: Utxo) => `${utxo.txid}:${utxo.vout}`;
+  const getOutpoint = (utxo: Utxo) => `${utxo?.txid}:${utxo?.vout}`;
```
Fahad-Mahmood, line 12: we can add this function to the context values as well. Looks like we are using it in a lot of components and then replace the duplicate code in UtxoItem and AccountUtxoList component
I'm not sure Outpoint is a good choice for this variable, maybe getUtxoKey

src/screens/AccountUtxoList/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { utxos } = currentAccount;
+
+  useEffect(() => {
+    navUtils.setHeaderTitle(currentAccount.name, navigation);
+  }, []);
+
+  const outpoint = (u: Utxo) => `${u.txid}:${u.vout}`;
```
Fahad-Mahmood, line 30: this is a duplicate function

src/screens/AccountUtxoList/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <View style={styles.container}>
+      <SelectedUtxosHeader toggleScreenAction="bubbles" navigation={navigation} />
```
Fahad-Mahmood, line 34: we can create a HeaderScreenActions enum and use it instead of "bubbles" string

src/components/accounts/SelectedUtxosHeader.tsx
```diff
@@ -0,0 +1,156 @@
...
+export default function SelectedUtxosHeader({
+  toggleScreenAction = 'bubbles',
+  navigation
+}: Props) {
+  const transactionBuilderContext = useTransactionBuilderContext();
+  const selectedUtxos = transactionBuilderContext.getInputs();
+  
+  const accountsContext = useAccountsContext();
+  const { currentAccount } = accountsContext;
+  const { utxos } = currentAccount;
```
Fahad-Mahmood, line 28: could be a single line destructure

src/components/accounts/SelectedUtxosHeader.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { useTransactionBuilderContext } from "./TransactionBuilderContext";
+import { AppText } from "../shared/AppText";
+import { Sats } from "./Sats";
+import { Colors, Typography } from "../../styles";
+import { Utxo } from "../../models/Utxo";
+import BubblesIcon from '../../assets/images/bubbles.svg';
+import ListIcon from '../../assets/images/list.svg';
+
+interface Props {
+  toggleScreenAction: 'bubbles' | 'list',
```
Fahad-Mahmood, line 15: this could be a an enum type or create a const object with these values so we don't have to explicitly type "bubbles"  in different screens and increase chances for typo errors.

Fahad-Mahmood (commented):
Looks good 👍 Added few comments!

src/screens/AccountUtxoList/index.tsx
tmakerman: `utxos` is a property of `Account`, not of `AccountsContext`

src/components/accounts/TransactionBuilderProvider.tsx
tmakerman: I've exposed and used the function through TransactionBuilderContext.

The spot where I'm rendering txid:vout in UtxoItem will be replaced with more UTXO details when I further work on the UTXO listing screen.

As to "Outpoint" name, I was just following what its called in bitcoin:
https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch06_transactions.adoc#outpoint
https://developer.bitcoin.org/reference/transactions.html#outpoint-the-specific-part-of-a-specific-output

src/screens/AccountUtxoList/index.tsx
tmakerman: I've exposed and used the function through TransactionBuilderContext for reuse.

src/screens/AccountUtxoList/index.tsx
tmakerman: I just learned you can destructure nested objects, so yes single line like this works:

`  const { currentAccount, currentAccount: { utxos } } = useAccountsContext();`

src/screens/AccountUtxoList/index.tsx
tmakerman: commenting on this one below...

src/components/accounts/SelectedUtxosHeader.tsx
tmakerman: I've thought this over.  The string literal union type should be as typesafe as an enum.  When you start typing the string the IDE should autocomplete the options.  In my testing if I set a value not in the literal union the IDE typescript service picks it up and complains.

Also interestingly neither the enum or literal union seem to be completely typesafe.  While the IDE flags it, I can still set values not in them and the app still runs without compile or runtime error.

The main reason I prefer the string literal union in this case is I don't need to expose an enum.  This header will be used from two screens only.  We could create a HeaderScreenActions enum as you mentioned above, but I wouldn't want to use it for actions that appear on multiple different screens - would want it confined to just actions of this component.

tmakerman:
Thanks for the review, @Fahad-Mahmood.  I'm done with my updates.  I could still be convinced on the enum thing 😉

src/components/accounts/SelectedUtxosHeader.tsx
Fahad-Mahmood: I was talking about string enums:

```
enum HeaderScreenActions {
 BUBBLES = 'bubbles',
 LIST = 'list
}
```

We are fine with string literal unions as well, if the usage is limited to one or two screens only.  
I guess we have to update the tsconfig file to avoid ignoring such errors during compilation.

Fahad-Mahmood (approved):
With the recent updates, this PR can be merged to production 👍

PR #41, opened 2024-05-28
Configure blockchain

tmakerman:
Introduces a "Configure Blockchain" screen allowing user to switch the network, backend protocol, and backend URL that will be connected to.  Currently supporting just testnet and signet.  Persists the selections (can change to mmkv after migration).  The dialog intentionally does not follow a design from Figma.  The Figma config screens will be implemented later.  This was implemented for testing purposes, though it will be useful for production use as well.


Psycarlo:
Looks good. Also tested on android.

PR #42, opened 2024-05-28
UTXO Bubbles Improvements

umarluqman:
This PR is to add improvements over the current zooming and panning for bubbles

- Double tap anywhere inside the canvas to zoom in into that particular area.
- Pinch anywhere inside the canvas instead from the center.
- Added decay animation when panning
- Bubble can move freely without zooming
- Integrate UTXO selection

Next
- Double tapping on any bubble to show details.
- Pinching any bubble to show details.

P/S: using entirely different code for gesture - most code from react native image zoom but with modifications to work with Skia and focal point zooming.


Psycarlo (approved):
A lot of complex code. Looks good for now

PR #43, opened 2024-06-07
Fixed Custom Font React Native Skia

umarluqman:
- Added custom font for react native skia
- Added UTXO details when zooming in

P/S: The font doesn't use expo-font because not possible in Skia canvas.


Psycarlo:
@umarluqman android fonts did not changed?
`mobile/android/app/src/main/assets/fonts`

PR #47, opened 2024-06-18
UTXO Bubbles Zoom Improvements

umarluqman:
- Show only the labels if zoomed radius (scale * radius) is more than 100.
- Fix flickering when double tapping (caveat: single tapping will have a slight delay)

Performance improvements
- Put useFonts on the parent instead of loading the fonts on each bubble.
- Memoize frequently render function


Psycarlo:
@umarluqman can you fix conflicts?
`apps/mobile/app/accountList/account/[id]/signAndSend/selectUtxoBubbles.tsx`
should now be
`apps/mobile/app/account/[id]/signAndSend/selectUtxoBubbles.tsx`

PR #52, opened 2024-12-17
docs: provide Android set up instructions

v4v2:
This provide detailed instructions to get up the application running, specially on Linux systems and also for developers using IDEs other than Android Studio.


Psycarlo (approved):

Psycarlo:
@v4v2 Thanks

PR #53, opened 2024-12-17
fix: unlock pin input not working

v4v2:
When the `handleOnFillEnded` is called, the state of the variable `pin` has not been updated yet even though `setPin` is called before `handleOnFillEnded`, so the value of pin is the previous one.

We add an optional parameter to `handleOnFillEnded`, which is the actual pin value itself, so when it is called it access this value instead of the outdated value stored in the state.


Psycarlo (approved):
Completed: https://trello.com/c/drq6PMT4/23-bug-pin-a-happycoder

PR #54, opened 2024-12-19
Transactions are clickable, takes you to a page to download bytes from Esplora and see results of Bdk calls for transaction

KitchenMishap:
1) On developer page, a button calls some methods in Bdk on the blockchain object, and gives some textual results below.
2) On Total Transactions list, transactions are now clickable and take you to a page for that transaction. On there, the txid is shown, and a button to fetch the transaction bytes from Esplora and shows the results of calling the methods in Bdk on the transaction object


Psycarlo:
Thanks **kitchen** for the PR. Happycoder is using some of your work on a new one.
Closing

PR #62, opened 2024-12-24
ui: focus on next word after finishing current (seed word input screen)

v4v2:


apps/mobile/components/SSWordInput.tsx
```diff
@@ -10,42 +10,53 @@ type SSWordInputProps = {
...
-  ...props
-}: SSWordInputProps) {
-  const textInputStyle = useMemo(() => {
-    return StyleSheet.compose(
-      {
-        ...styles.textInputBase,
-        ...(invalid ? styles.textInputInvalid : {})
-      },
-      style
+const SSWordInput = forwardRef(
```
Psycarlo, line 16: can we maintain the `function` keyword instead of arrow function, same as we doing for other components?

Psycarlo: `function SSWordInput`

`export default forwardRef(SSWordInput)`

this might work?

apps/mobile/app/(authenticated)/addMasterKey/importSeed.tsx
```diff
@@ -126,8 +134,9 @@ export default function ImportSeed() {
 
     setSeedWordsInfo(seedWords)
     setCurrentWordText(word)
-
-    // TODO: Set focus to next?
+    if (seedWord.valid) {
```
Psycarlo, line 137: small thing, but what about one-liner `if (seedWord.valid) focusNextWord(index)`?

Psycarlo:
@v4v2 When I type "win" but I want to write "winter", since win is a BIP39 word and "win " is seedWord.valid, it will focus on the next one without letting me to finish typing "winter"

v4v2:
> @v4v2 When I type "win" but I want to write "winter", since win is a BIP39 word and "win " is seedWord.valid, it will focus on the next one without letting me to finish typing "winter"

Ok, I will handle thsi edge case

Psycarlo (approved):

PR #83, opened 2025-01-10
fix bug paste

Jeezman:
- move clipbaord functionality to own file
- only paste if bitcoin address for now

### DIFF
--- a/apps/mobile/utils/clipboard.ts
+++ b/apps/mobile/utils/clipboard.ts
@@ -1,0 +1,35 @@
+/* eslint-disable no-console */
+import * as Clipboard from 'expo-clipboard'
+
+import { isBitcoinAddress } from './format'
+
+export async function setClipboard(value: string): Promise<void> {
+  try {
+    await Clipboard.setStringAsync(value)
+  } catch (error) {
+    console.error(error)
+  }
+}
+
+export async function getClipboard(): Promise<string | void> {
+  try {
+    if (!(await Clipboard.hasStringAsync())) {
+      return
+    }
+    const value = await Clipboard.getStringAsync()
+    if (!isBitcoinAddress(value)) {
+      return
+    }
+    return value
+  } catch (error) {
+    console.error(error)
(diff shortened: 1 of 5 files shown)


apps/mobile/utils/format.ts
```diff
@@ -56,11 +56,19 @@ function formatPageUrl(path: string, params: PageParams) {
   return url
 }
 
+// from https://stackoverflow.com/questions/21683680/regex-to-match-bitcoin-addresses + slightly modified to support testnet addresses
+function isBitcoinAddress(address: string) {
```
Psycarlo, line 60: I would not add this util inside `format`. Maybe create a new util file: `address.ts` or `bitcoin.ts`

apps/mobile/utils/clipboard.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-console */
```
Psycarlo, line 1: We don't want console logs.
For now we can just ignore the error
Just catch it and do nothing:

```
catch (_error) {
  //
}
```

apps/mobile/utils/format.ts
Psycarlo: Also, add some `unit` tests for this function under `tests/unit/utils`

Psycarlo (changes requested):

apps/mobile/utils/format.ts
Jeezman: fixed https://github.com/satsigner/satsigner/pull/83/commits/9f8670248e3992ab78a023a6df948794f43d6d6c

apps/mobile/utils/clipboard.ts
Jeezman: Got it. We should probably work towards adding an error logger. Fixed https://github.com/satsigner/satsigner/pull/83/commits/9f8670248e3992ab78a023a6df948794f43d6d6c

PR #85, opened 2025-01-11
Add bitcoin network

pedromvpg:


Jeezman (commented):
I think you need to update this file by adding `bitcoin` to this line https://github.com/satsigner/satsigner/blob/03446e4bbe1fc998092f9f78c620ed89af9e112f/apps/mobile/types/settings/blockchain.ts#L2

```
export type Network = 'testnet' | 'signet' | 'bitcoin'
```

Also, are you okay with calling it `bitcoin` instead of `mainnet`? @pedromvpg

Psycarlo:
@Jeezman We can change the label to whatever, but the value should be `bitcoin` because bdk accepts "bitcoin" and we are casting to bdk types. 
I should refactor this later.

Psycarlo (approved):

Psycarlo (approved):

PR #104, opened 2025-01-24
Toggle control for balance chart

Yi-Jacob:
Added toggle control for transaction balance chart.


apps/mobile/app/(authenticated)/settings/config/features/chartSetting.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+import SSButton from '@/components/SSButton'
+import SSCheckbox from '@/components/SSCheckbox'
+import SSText from '@/components/SSText'
+import SSMainLayout from '@/layouts/SSMainLayout'
+import SSVStack from '@/layouts/SSVStack'
+import { i18n } from '@/locales'
+import { useChartSettingStore } from '@/store/chartSetting'
+
+export default function ChartSettings() {
```
Psycarlo, line 14: Since we name the component ChartSettings and there are multiple settings, rename this file to chartSettings.tsx

apps/mobile/store/chartSetting.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type ChartSettingAction = {
+  setShowLabel: (showLabel: boolean) => void
+  setShowAmount: (showAmount: boolean) => void
+  setShowTransactionInfo: (showTransactionInfo: boolean) => void
+  setShowOutputField: (showOutputField: boolean) => void
+  setLockZoomToXAxis: (lockZoomToXAxis: boolean) => void
+}
+
+const useChartSettingStore = create<ChartSettingState & ChartSettingAction>()(
```
Psycarlo, line 22: Rename file to `chartSettings.ts`

Psycarlo: Also, in satsigner we will have multiples charts - not just Balance Chart. For now, we can leave the state like this, but in the future we might need to have settings for each chart.

Psycarlo (changes requested):

Psycarlo (approved):

PR #106, opened 2025-01-28
UI account settings view seed

v4v2:
This PR:

- adds function to decrypt account seed
- shows decrypted seed in account settings
- adds warning icon


apps/mobile/app/(authenticated)/account/[id]/settings/index.tsx
```diff
@@ -75,6 +79,14 @@ export default function AccountSettings() {
     router.replace('/')
   }
 
+  useEffect(() => {
+    const updateSeed = async () => {
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/account/[id]/settings/index.tsx:83]: use `function` keyword

Psycarlo (changes requested):

Psycarlo (approved):

PR #109, opened 2025-01-29
Feat skip pin dev settings

v4v2:
- adds SSSwitch component
- adds a skip pin switch on developer settings
- implements logic to skip pin

Currently, when SKIP PIN is ON, the Remember Last Visited Page feature shows a little warning but does not cause any error.


apps/mobile/app/(authenticated)/settings/config/developer.tsx
```diff
@@ -12,7 +14,11 @@ import { useAuthStore } from '@/store/auth'
 
 export default function Developer() {
   const deleteAccounts = useAccountsStore((state) => state.deleteAccounts)
-  const setFirstTime = useAuthStore((state) => state.setFirstTime)
+  const [setFirstTime, skipPin, setSkipPin] = useAuthStore((state) => [
```
Psycarlo, line 17: `useShallow` here

apps/mobile/app/(authenticated)/_layout.tsx
```diff
@@ -16,21 +16,46 @@ import type { PageRoute } from '@/types/navigation/page'
...
   const routeName = getFocusedRouteNameFromRoute(useRoute()) || ''
   const routeParams = useGlobalSearchParams()
 
   if (firstTime) return <Redirect href="/setPin" />
-  if (requiresAuth && lockTriggered) return <Redirect href="/unlock" />
+
+  if (requiresAuth && lockTriggered && !skipPin)
+    return <Redirect href="/unlock" />
+
+  if (lockTriggered && skipPin) {
```
Psycarlo, line 49: Is this `if` okay?

Restore app router stack only works if `skipPin` is true and `lockTriggered` is true?

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/_layout.tsx
v4v2: Yes because if skipPin is set to false it will redirect to unlock and the unlock.tsx already restores the router history after the screen is unlocked.

If skipn is set to true we need to restore the router history using a setTimeout because we cannot run `router.push` before the root layout is mounted, that is why we have a timeout.

Psycarlo: Got it.
But `setTimeout` here is not the best approach since mount can take more or less time.
Could there be a way to know when root is mounted and only then run the code...

Psycarlo (approved):

PR #122, opened 2025-02-08
feat: import watch-only wallet

v4v2:
- [X] add watch-only options
- [x] add descriptor input
- [x] add xpub input
- [x] add option to set master fingerprint
- [x] add option to set derivation path
- [x] validate input
- [x] auto-suggest derivation path from xpub
- [x] add tests for validation utils
- [x] create watch-only account from xpub
- [x] create watch-only account from descriptor
- [x] sync wallet (in other words, MAKE IT WORK)


Psycarlo:
I like the todo list ;)

Psycarlo (approved):

PR #126, opened 2025-02-13
Account Home Tab: Child Accounts

ericlee1112-dev:


apps/mobile/components/SSChildAccountList.tsx
```diff
@@ -0,0 +1,93 @@
+import { FlashList } from '@shopify/flash-list'
```
Psycarlo, line 1: No need to create a ChildAccountsList component. We can do this inline in the page?

Psycarlo (changes requested):

apps/mobile/components/SSChildAccountList.tsx
```diff
@@ -0,0 +1,93 @@
...
+            flex: 1
+          }}
+        >
+          {item.index}
+        </SSText>
+        <SSText style={{ color: '#fff', textAlign: 'center', flex: 2 }}>
+          {formatAddress(item.address, 4)}
+        </SSText>
+        <SSText
+          style={{
```
v4v2, line 49: please create a stylesheet using react-native `Stylesheet.create` to avoid repeating the style object

apps/mobile/app/(authenticated)/account/[id]/index.tsx
```diff
@@ -208,10 +218,207 @@ function TotalTransactions({
...
+              >
+                {type}
+              </SSText>
+            </SSHStack>
+          )
+        )}
+      </SSHStack>
+
+      <SSHStack
+        style={{
```
Comment (v4v2) [inline apps/mobile/app/(authenticated)/account/[id]/index.tsx:394]: consider putting the style literal objects in a `Stylesheet.create` stylesheet in order to keep things organized

Psycarlo (changes requested):

apps/mobile/components/SSChildAccountList.tsx
Psycarlo: Since its a page component and its a override we can leave inline @v4v2

apps/mobile/app/(authenticated)/account/[id]/index.tsx
Comment (Psycarlo) [reply apps/mobile/app/(authenticated)/account/[id]/index.tsx:394]: Since its a page component and its a override we can leave inline

apps/mobile/components/SSChildAccountList.tsx
Psycarlo: This will be in page component since we are removing the component

Psycarlo:
Known problems:
- Some styles in `app/(authenticated)/account/[id]/index.tsx` are created in `StyleSheet.create´ but only used once. Move them to inline.
- `fetchAddresses` being called 2 times after button click
- `wallet.getInternalAddress` not working correctly - showing same addresses as `wallet.getAddress`

Will be fixed in another PR

Psycarlo (approved):

PR #128, opened 2025-02-13
fix: utxo details i18n

v4v2:
- [x] fix translations in utxo details

### DIFF
--- a/apps/mobile/app/(authenticated)/account/[id]/transaction/[txid]/utxo/[vout]/index.tsx
+++ b/apps/mobile/app/(authenticated)/account/[id]/transaction/[txid]/utxo/[vout]/index.tsx
@@ -58,7 +58,7 @@ export default function UtxoDetails() {
     <ScrollView>
       <Stack.Screen
         options={{
-          headerTitle: () => <SSText>{t('utxoDetails.labelEdit')}</SSText>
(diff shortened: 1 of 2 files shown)


apps/mobile/app/(authenticated)/account/[id]/transaction/[txid]/utxo/[vout]/index.tsx
```diff
@@ -115,7 +111,7 @@ export default function UtxoDetails() {
           <SSClipboardCopy text={utxoAddress}>
             <SSVStack gap="none">
               <SSText weight="bold" uppercase>
-                {t('common.address')}
+                {t('transaction.address')}
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/account/[id]/transaction/[txid]/utxo/[vout]/index.tsx:114]: add duplicate in utxo key
`utxo.address`

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/account/[id]/transaction/[txid]/utxo/[vout]/index.tsx
Comment (Psycarlo) [reply apps/mobile/app/(authenticated)/account/[id]/transaction/[txid]/utxo/[vout]/index.tsx]: Fixed [d607ac3](https://github.com/satsigner/satsigner/pull/128/commits/d607ac300dd3fee12413da702688de1959368217)

PR #131, opened 2025-02-17
Network status & update styles

Yi-Jacob:
- Updated button layout
- Removed gradient background
- Display network status label
- Sync status automatically with settings and network status


apps/mobile/app/(authenticated)/index.tsx
```diff
@@ -57,19 +70,146 @@ export default function AccountList() {
...
+    if (
+      url === 'ssl://mempool.space:60602' ||
+      url === 'https://mutinynet.com/api'
+    ) {
+      return false
+    }
+    return true
+  }, [url])
+
+  const verifyConnection = useCallback(async () => {
```
Psycarlo, line 102: Maybe we can extract connection logic to custom hooks in `hooks` folder. What do you think?

apps/mobile/app/(authenticated)/index.tsx
```diff
@@ -57,19 +70,146 @@ export default function AccountList() {
...
+  const connectionString = useMemo(() => {
+    return network + ' - ' + url
+  }, [network, url])
+
+  const verifyUrl = useMemo(() => {
+    const urlObj = new URL(url)
+    if (urlObj.protocol === 'ssl:') {
+      const modifiedUrl = new URL(url.replace('ssl://', 'https://'))
+      modifiedUrl.port = ''
+      modifiedUrl.pathname = '/signet/api/v1/difficulty-adjustment'
```
Psycarlo, line 84: Hardcoded to signet. What happenes if we use mainnet or testnet?
This should be dynamic, right?

apps/mobile/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@bitcoinerlab/secp256k1": "^1.2.0",
+    "@react-native-community/netinfo": "^11.4.1",
```
Psycarlo, line 22: Add this dependency to `docs` - `dependencies.mdx` file

apps/mobile/styles/colors.ts
```diff
@@ -11,12 +11,15 @@ export const gray = {
   200: '#828282',
   300: '#777777',
   400: '#686868',
+  450: '#606060',
   500: '#515151',
   600: '#434343',
   700: '#363636',
   800: '#2D2D2D',
   850: '#242424',
+  870: '#212121',
```
Psycarlo, line 20: Values in the middle should be 875 and 925?

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/index.tsx
Yi-Jacob: Great idea.
Created custom hook `useVerifyConnection`

apps/mobile/app/(authenticated)/index.tsx
Yi-Jacob: Absolutely right. Fixed

apps/mobile/package.json
Yi-Jacob: Updated `dependencies.mdx`

apps/mobile/styles/colors.ts
Yi-Jacob: Fixed.

Psycarlo (approved):

Psycarlo:
Trello task for reference [here](https://trello.com/c/QYQ7QZZl/142-network-status-a-jacob)

PR #140, opened 2025-02-27
BIP48 support for multisig-wallet

Yi-Jacob:
- Each participant's derivation path follows BIP48 standard.
- Multi-sig wallet's descriptor now using sortedmulti from BIP67.
- Some UI restyling.


Psycarlo:
Closing. We will revisit this in the future.

PR #150, opened 2025-03-13
Feature: Converter / Currency Screen

NerdNook-rgb:
**In this Pull Request**

image

- [x] Converter / Currency routes: `currency/`
- [x] Implement the design
- [x] Create SSCurrencyInput component. (automatic  group separator, number format)
- [x] Use the mempool api to fetch prices
- [x] Sync inputted sats and bitcoin amounts.
- [x] Show the inputted bitcoin prices with different 6 fiat currencies


apps/mobile/components/SSCurrencyInput.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const rawValue = text.replace(/,/g, '')
+    if (/^(\d*\.?\d*)$/.test(rawValue) || rawValue === '') {
+      const formattedValue = formatNumberWithCommas(rawValue)
+
+      setLocalValue(formattedValue)
+      if (onChangeValue) onChangeValue(parseFloat(rawValue))
+    }
+  }
+
+  //   function handleSubmitText() {
```
Psycarlo, line 114: Keep the comment?

apps/mobile/store/price.ts
```diff
@@ -20,9 +20,8 @@ const SATS_IN_BITCOIN = 100_000_000
 const usePriceStore = create<PriceState & PriceAction>()((set, get) => ({
   prices: {
     AUD: 0,
-    BRL: 0,
```
Psycarlo, line 23: Why remove these?

apps/mobile/types/models/Blockchain.ts
```diff
@@ -37,15 +37,7 @@ export type PriceValue = {
   value: number
 }
 
-export type Currency =
```
Psycarlo, line 40: Why remove some?

apps/mobile/utils/format.ts
```diff
@@ -9,15 +9,30 @@ function formatAddress(address: string, character: number = 8) {
...
+  const formattedInteger = integerPart.replace(
+    /(\d)(?=(\d{3})+(?!\d))/g,
+    '$1' + separator
+  )
+
+  return decimalPart !== undefined
+    ? `${formattedInteger}.${decimalPart}`
+    : formattedInteger
+
+  // return formatted.replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1' + separator)
```
Psycarlo, line 35: Comment?

Psycarlo (changes requested):

apps/mobile/components/SSCurrencyInput.tsx
NerdNook-rgb: This function `handleSubmitText()` is not require for SSCurrencyInput component because there is no case the user need to be click "Done" or "Enter" from the keyboard. All the fiat currency prices automatically synced by input text change.

So, we can remove that function from the SSCurrencyInput component.
I will remove unnecessary codes.

apps/mobile/store/price.ts
NerdNook-rgb: In this basic version of Converter / Currency screen, we gather the prices from this api:
[](https://mempool.space/api/v1/prices)

This is response data from that api:
{"time":1741913102,"USD":81485,"EUR":74803,"GBP":62941,"CAD":117360,"CHF":71793,"AUD":129557,"JPY":12099834}

As you can see, they changed the default fiat currencies and that's why I changed the default currency list. In case we need other prices as well (it might be necessary in the next version), we should implement additional util function for that.

apps/mobile/types/models/Blockchain.ts
NerdNook-rgb: In this basic version of Converter / Currency screen, we gather the prices from this api:
[](https://mempool.space/api/v1/prices)

This is response data from that api:
{"time":1741913102,"USD":81485,"EUR":74803,"GBP":62941,"CAD":117360,"CHF":71793,"AUD":129557,"JPY":12099834}

As you can see, they changed the default fiat currencies and that's why I changed the default currency list. In case we need other prices as well (it might be necessary in the next version), we should implement additional util function for that.

apps/mobile/utils/format.ts
NerdNook-rgb: That commented code is unnecessary snippet anymore. I've implemented the return feature above on that line. I will remove that one.

Psycarlo (approved):

francismars (approved):
ack

PR #151, opened 2025-03-14
Display that Accounts are Syncing

Psycarlo:
- [x] Account model isSyncing type
- [x] User feedback when the account is syncing


francismars (approved):
Ack

PR #163, opened 2025-03-21
refactor: use flashList

Jeezman:
### What does PR do?

Task is to refactor FlatList to FlashList. 

https://trello.com/c/NUSr2Nwh/145-refactor-flashlist-on-all-lists-a-tobi


Jeezman:
@Psycarlo kindly review

Psycarlo:
Hi @Jeezman 
Based on this [trello task](https://trello.com/c/NUSr2Nwh/145-refactor-flashlist-on-all-lists-a-tobi)
For `lists` I meant `"for loops"`
No need to refactor the SSKeyboardWordSelector because we know it won't have many words, just a few to pick from.

We want to introduce FlashList in the account transations list, for example:

`account/[id]/index.tsx`

```tsx
{sortedTransactions
  .slice()
  .reverse()
  .map((transaction, index) => {
...
})}
```

☝🏼 Introduce Flash List in these cases.

**Where to use FlashList?**
- Account List (user can have many accounts? [more than 10])
- Account Transactions List
- Account UTXO List

I think that's it for now.

Psycarlo:
Also check this bug: - _FlashList's rendered size is not usable._

Psycarlo (approved):

PR #167, opened 2025-03-21
Keep demo account block after key creation

pedromvpg:
- [x] Keep samples after first key is created
- [x] Change button styles
- [x] Add title

image


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/accountList.tsx
```diff
@@ -263,6 +232,40 @@ export default function AccountList() {
               </SSVStack>
             ))}
           </SSVStack>
+          <SSVStack itemsCenter style={{ paddingTop: 50, paddingBottom: 100 }}>
```
Psycarlo, line 235: Using padding bottom here is not ideal.

Psycarlo (changes requested):

Psycarlo (approved):

PR #178, opened 2025-04-01
Transactions Refine

Psycarlo:
## Transactions Refine

In this PR we introduce missing functionality to the transactions follow

### Added functionality

- [ ] Inputs selection: User, Privacy & Efficiency [Requires @Jeezman to refactor algos ReturnType's]
- [x] Require outputs to run Privacy and Efficiency algorithms
- [x] ~~Timelock~~ (`nTimeLock` not present: https://github.com/LtbLightning/bdk-rn/blob/main/lib/classes/TxBuilder.d.ts)
- [x] Allow Replace by fee
- [x] ~~Child pays for parent~~ Always possible - does not need to be activated
- [x] Sankey: Show Labels @umarluqman 
- [x] Sankey: Edit/Update Outputs
- [x] Sankey: Update based on current selection
- [ ] Sankey: Miner fee (initial selected fee (estimate) and current fee) [use RECOMMENDED_BASE_FEE]
- [x] Output Bottom Sheet: Remove Output
- [x] Change Fee Bottom Sheet: Display Fee Chart [Needs improvements from @Jeezman]
- [x] Change Fee Bottom Sheet: Sats input via numpad ** [`SSNumberGhostInput`]
- [x] Output Bottom Sheet: Sats input via numpad ** [`SSNumberGhostInput`]
- [x] Click on minerFee output opens change fee bottom sheet 
- [x] Click on change output opens add output bottom sheet 
- [x] Require above dust limit (Hardcode to 546)
- [x] Scan (camera): Allow & populate BIP21 (navigates to selectUtxoList)*
- [x] Validation: Valid bitcoin address, Valid BIP21 & Valid bip21 amount <= remainingSats
- [ ] Validation: Missing keys
- [x] Warning and error toasts
- [x] Sign Transaction UI: message id, sankey, full message @v4v2 
- [x] Signed Transaction UI @v4v2
- [x] Broadcasted Transaction UI @v4v2

### Fixes

- [x] Padding right for `SSTextInput` if `actionRight` is present
- [x] Refactor: create ellipsis icon
- [x] Add change icon and text to change output in Sankey Diagram
- [x] Remove TouchableOpacity from Ellipsis Icon
- [x] Remove generateId, replace with uuid 
- [x] Refactor `SATS_PER_BITCOIN` - removed repeated instances - place it inside constants folder
- [x] Used SF Pro Text in SSBubble for Android @umarluqman  
- [x] Fix SSFeeInput not displaying correct information***

### Improvements

- [x] ~~Lazy load bottom sheets~~ To be done in another PR
- [x] Loading state when loading Sankey Diagram

### Warnings & Erros

- [x] Display errors on build transaction
- [x] Display toast error on wrong bitcoin address (scan qr code)
- [x] Display toast warning on BIP21 amount > remainingSats

### Design
<details>
<summary>Click to reveal figma design</summary>

image

image

image

image
</details>

### Notes

(*) When scanning, it can be a bitcoin address or bip21. Does not make sense to navigate to selectUtxoList because we user still needs to add the amount and label if not present in address/bip21. Also, if we navigate to selectUtxoList after scanning, user does not have feedback that the uxto was added. Even if we display a toast, we need to add the output to the transaction builder store with an arbritary amount and label that the user would need to change.
** Still needs to show suffix during editing. Add this logic to the SSNumberInput, with an optional prop `suffix`
*** Still need to think about what should be the max value and how we should ~estimate the block


francismars (approved):
ack

PR #180, opened 2025-04-02
feat: Enter pin to reveal mnemonic

Jeezman:
## What does this PR do?

Enables PIN entry before mnemonic is revealed


apps/mobile/components/SSPinEntry.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useShallow } from 'zustand/react/shallow'
+
+import SSPinInput from '@/components/SSPinInput'
+import SSText from '@/components/SSText'
+import { useAnimatedShake } from '@/hooks/useAnimatedShake'
+import SSVStack from '@/layouts/SSVStack'
+import { t } from '@/locales'
+import { useAuthStore } from '@/store/auth'
+
+interface SSPinEntryProps {
```
Psycarlo, line 12: use `type` instead of `interface`

apps/mobile/components/SSPinEntry.tsx
```diff
@@ -0,0 +1,54 @@
...
+  pin: string[]
+  setPin: Dispatch<SetStateAction<string[]>>
+  onFillEnded: (pin: string) => void
+  title?: string
+}
+
+export default function SSPinEntry({
```
Psycarlo, line 19: export default on the bottom of the component to be consistency with the other components

Psycarlo (changes requested):

apps/mobile/components/SSPinEntry.tsx
Jeezman: fixed https://github.com/satsigner/satsigner/pull/180/commits/a4a02aa34d30f9c1998d04b27773b2a27c4c6f05

Psycarlo:
On my physical device, enter pin modal does not trigger the keyboard immediately and entering the pin on the pin modal is laggy

Jeezman:
@Psycarlo Please do a screen record if possible so I can observe the lag. Does it also do same in the unlock screen? Same component is used there

Psycarlo (approved):

PR #181, opened 2025-04-03
Type update

Jeezman:
Update type and document effectiveValue

### DIFF
--- a/apps/mobile/types/models/Utxo.ts
+++ b/apps/mobile/types/models/Utxo.ts
@@ -7,6 +7,4 @@ export type Utxo = {
   addressTo?: string
   keychain: 'internal' | 'external'
   script?: number[]
-  effectiveValue?: number
-  scriptType?: 'p2pkh' | 'p2wpkh' | 'p2sh-p2wpkh'
 }
--- a/apps/mobile/types/models/Utxo.ts
+++ b/apps/mobile/types/models/Utxo.ts
@@ -7,4 +7,6 @@ export type Utxo = {
   addressTo?: string
   keychain: 'internal' | 'external'
   script?: number[]
+  effectiveValue?: number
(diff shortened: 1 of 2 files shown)


apps/mobile/types/models/Utxo.ts
```diff
@@ -7,4 +7,6 @@ export type Utxo = {
   addressTo?: string
   keychain: 'internal' | 'external'
   script?: number[]
+  effectiveValue?: number
```
Psycarlo, line 10: Don't add effectiveValue and scriptType to Utxo model

apps/mobile/utils/utxo.ts
```diff
@@ -1,10 +1,5 @@
 import { type Utxo } from '@/types/models/Utxo'
 
-type _Utxo = Utxo & {
```
Psycarlo, line 3: We can leave like this for the functions arguments
`utxos: _Utxo[]`

apps/mobile/utils/utxo.ts
```diff
@@ -30,7 +25,7 @@ type UtxoOptions = {
  * @returns {Object} Selected UTXOs and change
  */
 function selectEfficientUtxos(
```
Psycarlo, line 27: `selectEfficientUtxos` still returns
```
 {
    inputs: any;
    fee: number;
    change: number;
} | {
    inputs: never[];
    fee: number;
    change: number;
    error: string;
}
```

inputs type should be `Utxo[]` (and not _Utxo[])
i.e., effictiveValue and scriptType are only going to be used inside the algorithms and will not be stored in the Utxo model

Psycarlo (changes requested):

apps/mobile/utils/utxo.ts
Psycarlo: @Jeezman the return type still any

PR #186, opened 2025-04-06
Styles-and-language

pedromvpg:
Standardised styles of some headers.
Changed mono-space typeface to use the same family as the rest of the app: SF.
Removed field separator on decoded transaction
Reintroduced color coding on decoded transaction (🩸red for input data, 🌱 green for output data)
Added zoom in/out function to decoded transaction
Restyled decoded transaction transcription
Enabled deselection of field

image
image

ezgif-80260ac3e1896c


Psycarlo:
@pedromvpg CI is failing
Also, add a PR description

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/settings/import/labels.tsx
```diff
@@ -93,12 +94,19 @@ export default function ImportLabels() {
     <ScrollView style={{ width: '100%' }}>
       <Stack.Screen
         options={{
-          headerTitle: () => <SSText size="xl">{t('settings.title')}</SSText>,
+          headerTitle: () => (
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/settings/import/labels.tsx:97]: We are always adding this to headerTitle on these screens
We better this override on a more-top-level `_layout.tsx`

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/settings/index.tsx
```diff
@@ -401,23 +388,35 @@ export default function AccountSettings() {
...
+                      </SSHStack>
                     </View>
                   ))}
                 </SSSeedLayout>
               )}
             </SSHStack>
-            <SSTextClipboard text={localMnemonic.replaceAll(',', ' ')}>
-              <SSButton label={t('common.copy')} />
-            </SSTextClipboard>
+            <SSHStack>
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/settings/index.tsx:415]: Why introduce an additional `SSHStack`?

apps/mobile/components/SSModal.tsx
```diff
@@ -9,13 +9,24 @@ import SSButton from './SSButton'
 type SSModalProps = {
   visible: boolean
   fullOpacity?: boolean
+  variant?:
```
Psycarlo, line 12: This is not the modal "variant" but the close button "variant"
We maybe should change the prop name to `closeButtonVariant`
This variant should use `SSButtonProps['variant']` type (you need to export SSButtonProps type)

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/settings/index.tsx
Comment (v4v2) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/settings/index.tsx]: solved

apps/mobile/components/SSModal.tsx
v4v2: solved

Psycarlo (approved):

PR #189, opened 2025-04-07
Fix: Fixed Tab navigation issues

NerdNook-rgb:
In this pull request

I've fixed following issues
- [ ] Back arrow icon appear on tab's home screen after delete account
- [ ] Not able to back to tab's home screen in some case

### DIFF
--- a/apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
+++ b/apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
@@ -1,6 +1,7 @@
-import { Tabs, usePathname, useSegments } from 'expo-router'
+import { Tabs, usePathname, useRouter, useSegments } from 'expo-router'
 import { useEffect, useState } from 'react'
 import { StyleSheet, View } from 'react-native'
+import { TouchableWithoutFeedback } from 'react-native-gesture-handler'
 
 import {
   SSIconConverter,
@@ -16,8 +17,30 @@ import { showNavigation } from '@/utils/navigation'
 
 export default function TabLayout() {
   const currentPath = usePathname()
-  const segments = useSegments()
+  const router = useRouter()
+  const segments = useSegments() as string[]
   const [isShowTab, setShowTab] = useState(false)
+
+  const renderTabButton = (props: any, segment: string) => {
+    return (
+      <View style={props.style}>
+        <TouchableWithoutFeedback
+          onPress={() => {
+            if (
+              segments.indexOf(segment) >= 0 &&
+              segments.indexOf(segment) < segments.length - 1
+            ) {
+              router.replace(`/(authenticated)/(tabs)/${segment}`)
+            } else {
+              props.onPress?.()
+            }
+          }}
+        >
+          {props.children}
+        </TouchableWithoutFeedback>
+      </View>
+    )
+  }
 
   useEffect(() => {
     setShowTab(showNavigation(currentPath, segments.length))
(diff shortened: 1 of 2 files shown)


NerdNook-rgb:
image

apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
```diff
@@ -16,9 +17,31 @@ import { showNavigation } from '@/utils/navigation'
...
-  const segments = useSegments()
+  const router = useRouter()
+  const segments = useSegments() as string[]
   const [isShowTab, setShowTab] = useState(false)
 
+  const renderTabButton = (props: any, segment: string) => {
+    return (
+      <View style={props.style}>
+        <TouchableWithoutFeedback
+          onPress={() => {
```
v4v2, line 28: please move the function to outside the JSX, below the useState

v4v2 (changes requested):

apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
NerdNook-rgb: Thanks for your review and suggestion.

This is fixed code

```
  function handleTabItemPress(props: any, segment: string) {
    if (
      segments.indexOf(segment) >= 0 &&
      segments.indexOf(segment) < segments.length - 1
    ) {
      router.replace(`/(authenticated)/(tabs)/${segment}`)
    } else {
      props.onPress?.()
    }
  }

  const renderTabButton = (props: any, segment: string) => {
    return (
      <View style={props.style}>
        <TouchableWithoutFeedback
          onPress={() => handleTabItemPress(props, segment)}
        >
          {props.children}
        </TouchableWithoutFeedback>
      </View>
    )
  }
```

Psycarlo: @NerdNook-rgb don't use any

apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
```diff
@@ -27,8 +27,10 @@ export default function TabLayout() {
       segments.indexOf(segment) >= 0 &&
       segments.indexOf(segment) < segments.length - 1
     ) {
+      // @ts-ignore
```
Psycarlo, line 30: Why use ts-ignore?
We should avoid it at all costs

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
```diff
@@ -37,29 +65,33 @@ export default function TabLayout() {
...
+        backBehavior="history"
       >
         <Tabs.Screen
           name="(signer)"
           options={{
             title: 'Signer',
             tabBarIcon: ({ focused }) =>
-              renderTabIcon(focused, SSIconSignerActive, SSIconSigner)
+              renderTabIcon(focused, SSIconSignerActive, SSIconSigner),
+            tabBarButton: (props) => renderTabButton(props, '(signer)')
```
Psycarlo, line 76: Also, can you explain why you added `tabBarButton`

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/(tabs)/_layout.tsx
NerdNook-rgb: fixed

Psycarlo (approved):

PR #193, opened 2025-04-09
Feat/spiral gestures

umarluqman:
- [x] added gestured using useGestures hook
- [x] use custome font correctly
- [x] performance improvement


apps/mobile/components/SSSpiralBlocks.tsx
```diff
@@ -367,4 +432,14 @@ const styles = StyleSheet.create({
   }
 })
 
-export default SSSpiralBlocks
+// Use React.memo to prevent unnecessary re-renders
+export default React.memo(SSSpiralBlocks, (prevProps, nextProps) => {
```
Psycarlo, line 436: Can't we just use memo, instead React.memo?
Import `memo` from React

Psycarlo (changes requested):

Psycarlo (approved):

PR #199, opened 2025-04-23
NFC

pedromvpg:
Add NFC support and import descriptors with NFC on Watch only. Todo: Multisig

Enabled pre-existing NFC button and populate text field with scanned data. Validate for descriptors.

Animate button while NFC scan is active.


apps/mobile/utils/validation.ts
```diff
@@ -18,11 +18,66 @@ export function validateFingerprint(fingerprint: string) {
...
-    /^(sh|wsh|pk|pkh|wpkh|combo|multi|sortedmulti|tr|addr|raw|rawtr)\((\[([a-fA-F0-9]{8})?(\/[0-9]+[h']?)+\])?[a-z0-9]+(\/[0-9*])*\)(#[a-z0-9]{8})?$/gim
+  // Basic descriptor format validation
+  const basicRegex = new RegExp(
+    /^(sh|wsh|pk|pkh|wpkh|combo|multi|sortedmulti|tr|addr|raw|rawtr)(\((\[([a-fA-F0-9]{8})?(\/[0-9]+[h']?)+\])?[a-z0-9]+(\/[0-9*])*\))?(\/[0-9*])*(#[a-z0-9]{8})?$/gim
   )
-  return descriptor.match(r) !== null
+  basicRegex.lastIndex = 0
+
+  // Special handling for nested descriptors with xpub keys
+  if (descriptor.startsWith('sh') || descriptor.startsWith('wsh')) {
```
v4v2, line 28: Hi Pedro, all of the logic I highlighted is already checked by the rgex. The basicRegex is not just basic, it does validates public keys, fingerprints and checksums. Therefore, this code with several if blocks is not necessary, it is already validated by the regex

apps/mobile/utils/validation.ts
```diff
@@ -18,11 +18,66 @@ export function validateFingerprint(fingerprint: string) {
...
 export function validateDescriptor(descriptor: string) {
-  // TODO: this validates simple descriptors, but not complex ones
-  const r = new RegExp(
-    /^(sh|wsh|pk|pkh|wpkh|combo|multi|sortedmulti|tr|addr|raw|rawtr)\((\[([a-fA-F0-9]{8})?(\/[0-9]+[h']?)+\])?[a-z0-9]+(\/[0-9*])*\)(#[a-z0-9]{8})?$/gim
+  // Basic descriptor format validation
+  const basicRegex = new RegExp(
+    /^(sh|wsh|pk|pkh|wpkh|combo|multi|sortedmulti|tr|addr|raw|rawtr)(\((\[([a-fA-F0-9]{8})?(\/[0-9]+[h']?)+\])?[a-z0-9]+(\/[0-9*])*\))?(\/[0-9*])*(#[a-z0-9]{8})?$/gim
   )
-  return descriptor.match(r) !== null
+  basicRegex.lastIndex = 0
```
v4v2, line 25: Hi Pedro, all of the logic I highlighted is already checked by the rgex. The basicRegex is not just basic, it does validates public keys, fingerprints and checksums. Therefore, this code with several if blocks is not necessary, it is already validated by the regex

apps/mobile/utils/validation.ts
```diff
@@ -18,11 +18,66 @@ export function validateFingerprint(fingerprint: string) {
...
+  if (descriptor.includes('#')) {
+    const [desc, checksum] = descriptor.split('#')
+    if (!desc || !checksum || checksum.length !== 8) {
+      return false
+    }
+    return basicRegex.test(desc)
+  }
+
+  // Handle basic descriptors
+  return basicRegex.test(descriptor)
```
v4v2, line 80: Hi Pedro. All of the logic added is actually already checked by the rgex. The basicRegex is not just basic, it does validates public keys, fingerprints and checksums. Therefore, this code with several if blocks is not necessary, it is already validated by the regex. We can safely remove it.

You can also take a look at the unit tests and realize that this regex is enough to pass the tests cases for several scenarios

v4v2 (changes requested):

Psycarlo (approved):

PR #204, opened 2025-05-01
performance watch-only 

v4v2:
# Global

- [x] add state variables to track account sync progress

## Watch-Only Address wallet

- [x] refactor hooks used to fetch data
- [x] lazy load transactions
- [x] lazy load utxos
- [x] implement logic to prevent fetching the same transaction twice
- [x] implement logic to prevent fetching the same utxo twice
- [x] implement logic to track fetching progress
- [x] add UI progress status
- [x] load account summary (balance, sats in mempool, total txs, total utxos) as soon as possible
- [x] load transactions one by one (esplora)
- [x] load transactions one by one (electrum)

Final check:

- [x] make sure it works
- [x] make sure there is no bug updating account in the store
- [x] make sure the account is partially updated if something fails during the syncing (Esplora)
- [x] make sure the account is partially updated if something fails during the syncing (Electrum)

## FOR THE ~FUTURE~ PRESENT

Enhancements ~not~ included in this pull request ~but which shall be done in a smaller future pull request~.

- [x] Ensure Esplora client fetch all transactions not just the first 50 (for example, Mempool's Esplora API only returns 50 transactions per request which requires making multiple requests if we have hundreds of transactions)
- [x] Make Electrum client parse all transactions, old and news, when computing the income and outgoing value, as well as inputs and outputs


apps/mobile/hooks/useSyncAccountWithAddress.ts
```diff
@@ -24,6 +41,278 @@ function useSyncAccountWithAddress() {
...
+          script: parseHexToBytes(out.scriptpubkey)
+        })
+        if (out.scriptpubkey_address === address) {
+          received += out.value
+        }
+      })
+
+      const raw = await esploraClient.getTxHex(t.txid)
+
+      const tx = {
```
v4v2, line 127: change type

apps/mobile/hooks/useSyncAccountWithAddress.ts
```diff
@@ -24,6 +41,278 @@ function useSyncAccountWithAddress() {
...
+
+        let script: number[] | undefined
+
+        if (txDictionary[u.txid] !== undefined) {
+          const index = txDictionary[u.txid]
+          const tx = esploraTxs[index]
+          script = parseHexToBytes(tx.vout[u.vout].scriptpubkey)
+        }
+
+        return {
```
v4v2, line 169: add type

v4v2:
Actually, I will include in this PR the enhancements which before were meant to be included in the future

apps/mobile/api/esplora.ts
```diff
@@ -160,7 +160,25 @@ export default class Esplora {
   }
 
   async getAddressTx(address: string) {
```
Psycarlo, line 162: Should be `getAddressTxs` ?

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/index.tsx
```diff
@@ -646,12 +647,18 @@ export default function AccountView() {
...
-      state.accounts.find((account) => account.id === id),
-      state.updateAccount
-    ])
-  )
+  const [account, updateAccount, totalTasks, tasksDone, syncStatus] =
+    useAccountsStore(
+      useShallow((state) => [
+        state.accounts.find((account) => account.id === id),
+        state.updateAccount,
+        state.accounts.find((account) => account.id === id)?.syncProgress
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/index.tsx:655]: Is there a way to refactor this to not do 4 `.find()`?

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/index.tsx
Comment (Psycarlo) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/index.tsx:655]: Maybe two `useWalletsStore`. First to select `wallet and updateAccount` and the other for `syncProgress`, etc.?

Comment (v4v2) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/index.tsx:655]: > Is there a way to refactor this to not do 4 .find()?

I could not figure it out because accessing it from the store would make the nested object not reactive

PR #205, opened 2025-05-06
Feature: Parse multiple outputs format

NerdNook-rgb:
In this pull request

image

✅ Implementation:
- [x] Parse multiple outputs format

### DIFF
--- a/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/signAndSend/importOutputs.tsx
+++ b/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/signAndSend/importOutputs.tsx
@@ -21,9 +21,7 @@ function ImportOuputs() {
   const [importedOutputs, setImportedOutputs] = useState(tn('emptyContent'))
   const [validInput, setValidInput] = useState(false)
 
-  const [addOutput] = useTransactionBuilderStore(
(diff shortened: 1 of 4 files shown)


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/signAndSend/importOutputs.tsx
```diff
@@ -18,13 +21,22 @@ function ImportOuputs() {
   const [importedOutputs, setImportedOutputs] = useState(tn('emptyContent'))
   const [validInput, setValidInput] = useState(false)
 
+  const [addOutput] = useTransactionBuilderStore(
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/signAndSend/importOutputs.tsx:24]: Since we are only using one piece of state, we don't need an `array` or `useShallow`

`const addOutput = use...`

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/signAndSend/importOutputs.tsx
Comment (NerdNook-rgb) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/account/[id]/signAndSend/importOutputs.tsx]: I've removed array and useShallow

const addOutput = useTransactionBuilderStore((state) => state.addOutput)

Psycarlo (approved):

PR #211, opened 2025-05-12
Duress PIN

v4v2:
This pull request:

- [x] adds logic for duress pin
- [x] deletes all accounts if user inputs the duress pin
- [x] adds UI for setting duress pin
- [x] make sure accounts are not deleted if duress pin is not set
- [x] delete evidence that duress existed: if duress is ever entered, the standard pin is set to the duress pin and the duress pin is deleted
- [x] handle errors, ensure it works
- [x] ensure duress pin and pin are distinct
- [x] add modal to inform users about the duress functionality


apps/mobile/app/unlock.tsx
```diff
@@ -47,8 +47,26 @@ export default function Unlock() {
...
+    }
+
     const encryptedPin = await pbkdf2Encrypt(pin, salt)
-    const isPinValid = encryptedPin === storedEncryptedPin
+    const isPinValid =
+      encryptedPin === storedEncryptedPin ||
+      encryptedPin === storedEncryptedDuressPin
+
+    if (encryptedPin === storedEncryptedDuressPin) {
+      deleteAccounts()
```
Psycarlo, line 63: deleteWallets as well, from wallets store
Also we must delete tags from accounts store

Psycarlo (changes requested):

apps/mobile/app/unlock.tsx
v4v2: thanks for remembering @Psycarlo

Psycarlo (approved):

PR #219, opened 2025-05-21
chore: remove non-functioning public electrum mainnet servers

v4v2:
The removed servers were not working by the date of the commit.


Psycarlo:
@v4v2 Let's not ignore it. Skip the test or don't include it.

Psycarlo (approved):

PR #250, opened 2025-08-28
Remove package-lock

Psycarlo:
- The `package-lock.json` file is for `npm` and we are using `yarn`
- Remove additional satsigner ios project added by  @pedromvpg , by mistake
- Update pods


francismars:
ack

francismars (approved):
ack

PR #252, opened 2025-09-01
multiple language support for bip39 word list + fix madness with descriptor, fingerprint and extended key derivation

v4v2:
The goal of this pull request is to provide multi-language support for BIP 39.

- [x] adds interface to select bip39 word list
- [x] adds UI option to select default word list language
- [x] adds UI option to pick word list language when creating new wallet
- [x] shows word list according to the selected language in the UI
- [x] generates mnemonic to the select word list (single signature)
- [x] imports mnemonic from the select word list (single signature)
- [x] unit tests for converting mnemonic between bip39 word lists
- [x] mnemonics addresses and root keys match https://iancoleman.io/bip39/

Extra:

- [x] fixes mnemonic utilities in utils/bip39 and utils/bip32
- [x] unit tests for bip32 descriptors
- [x] unit tests for bip32 derivation of extended keys
- [x] fix descriptor derivation from mnemonic
- [x] fix extended key derivation from mnemonic
- [x] fix fingerprint derivation from mnemonic
- [x] fix descriptor derivation from extended key

FINAL TEST

- [x] PASS SEND/RECEIVE FUNDS FROM WALLETS

For a future pull request: support for multi-sig wallets to generate and import mnemonic for other word lists. Currently only supporting single signature.


apps/mobile/utils/bip39.ts
```diff
@@ -1,8 +1,79 @@
 import { getDefaultWordlist, wordlists } from 'bip39'
 
-function getWordList() {
-  const name = getDefaultWordlist()
+type WordList =
```
Psycarlo, line 3: Type can be infer from the `WORDLIST_LIST` array:

`const WORDLIST_LIST = [...] as const`
`type WordList = typeof WORDLIST_LIST[number]`

Psycarlo (changes requested):

v4v2:
still need to fix a bug.... not ready to be merged but soon will

PR #257, opened 2025-09-11
Add option in security settings to change pin

v4v2:
This pull request:

- [x] improves UI of `Settings > Security` screen
- [x] adds option under `Settings > Security` to change the current PIN
- [x] fixes bug where pressing to set pin later would erase current PIN (
  happened when canceling setting the pin for first time under `Developer
  Settings`)
- [x] re-encrypts all account secrets before setting new PIN
- [x] FINAL TEST (IS WORKING)

### AFTER

<hr />


apps/mobile/app/setPin.tsx
```diff
@@ -39,16 +49,73 @@ export default function SetPin() {
     confirmationPinArray.findIndex((text) => text === '') === -1
   const pinsMatch = pinArray.join('') === confirmationPinArray.join('')
 
+  // for each account, update re-encrypt each of its secret with new PIN
+  async function reEncryptAccounts(
```
Psycarlo, line 53: Should it be hook?

Psycarlo (changes requested):

Psycarlo (approved):

PR #328, opened 2026-03-25
Prevent auto display of large tx

v4v2:
- [X] Transaction Chart
- [X] Current Transaction Chart
- [x] Transaction decoded
- [X] Script decoded
- [x] Transaction input list
- [x] Transaction output list
- [x] Add generic component to avoid duplicate logic (it is awesome !!!)
- [x] Manually test it works


Psycarlo:
@v4v2 This is temporary only. It's bad UX. We need to find a way to circumvent this.

Psycarlo (approved):

PR #336, opened 2026-04-06
enhance Nostr Sync "trust devices"

v4v2:
When the user clicks to "trust device" on Nostr Sync, the following modal will be shown:

Screenshot_2026040901

------

## OLD description

Anybody can create a watch-only address for a given address (or multiple), thus an attacker can create watch-only address wallet with someone else address and potentially be able to retrieve labels from the legimate user of the address, and maybe trick him with social engineering through NostrSync Chat.

While a diligent user will never click "Trust device" of an unknown device, a naive/newbie user may accidentally (or by curiosity) click to "Trust device" belonging to an attacker. The attacker will see the user has trusted his device, and will start sharing the labels (utxos, transactions) from the watch-only address wallet, invading privacy of that user. The attacker can then start a chat via NostrSync Chat and pretend to have access to the user's fund and make threats and social engineering tactics -- and the fact that the labels are exposed will help the attacker to device more specific and sophisticated attacks based upon information extracted from labels.

While it makes sense to have NostrSync for common wallets, or watch-only wallets based upon public descriptor (shared only with trusted parties) and public kes (shared only with trusted parties), it is dangerous to have NostrSync for watch-only addresses, since the address is publicly listed on blockchain (the descriptor and xpub are not published on blockchain and must be manually shared beforehand).

@pedromvpg your thoughts.


v4v2:
due to @pedromvpg comments, I decide to not disable Nostr Sync but just to add a warning before trusting the device.

PR #338, opened 2026-04-12
Security enhancement (nostr sync): show warning about implications of trusting new devices

v4v2:
When the user clicks to "trust device" on Nostr Sync, the following modal will be shown:

Screenshot_2026040901

This is to make sure the user knows the implications of trusting new devices.


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/[id]/settings/nostr/index.tsx
```diff
@@ -661,6 +674,17 @@ export default function NostrSync() {
     ]
   )
 
+  const handleToggleMember = useCallback(
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/[id]/settings/nostr/index.tsx]: @v4v2 

Now that we are using react-compiler, we don't need `useCallback`

The compiler will handle memoization if necessary

Psycarlo (changes requested):

Psycarlo (approved):

PR #346, opened 2026-04-18
fix: not validating pin in multisig wallet view mnemonic page (SEVERE)

v4v2:
the pin is not validated before seed decryption. Any pin bypass authentication.


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/add/multiSig/export/seedWords.tsx
```diff
@@ -92,8 +92,22 @@ export default function SeedWordsPage() {
     }
   }, [accountData, key])
 
-  async function handlePinEntry() {
-    await decryptMnemonic()
+  async function handlePinEntry(pinString: string) {
```
Psycarlo, line 95: Same function as in the single seed export. Should we move this to the util folder?

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/add/multiSig/export/seedWords.tsx
v4v2: this is meant to be a hotfix. I was thinking later wrapping up everything in a new component that would have this validation logic (comparing with auth pin). Wanna fix in this current PR? @Psycarlo

Psycarlo: Let's do it fast here @v4v2

Psycarlo (approved):

PR #352, opened 2026-04-23
CI build: remove hard-coded project ID + update docs

v4v2:


apps/docs/content/docs/getting-started/build.mdx
```diff
@@ -3,17 +3,41 @@ title: Build
...
-2. Install `eas` on your machine
+2. Log in into your Expo account and create a new project
+
+3. Set `EXPO_PROJECT_ID` to the project id created:
+
+```shell
+export EXPO_PROJECT_ID='rpr6osn4-r915-4np9-ns31-q3724o78r59s'
+```
+
+Alternatively, hardcode it (do not commit) the file `apps/mobile/app.config.ts`:
```
Psycarlo: The file needs to be commited. What doesn't is the projectId line

Psycarlo (changes requested):

apps/docs/content/docs/getting-started/build.mdx
v4v2: what I meant here is if the dev is having trouble exporting `EXPO_PROJECT_ID`, he can temporarily hardcode the value in that file, but not commit this change.

apps/mobile/app.config.ts
```diff
@@ -1,5 +1,7 @@
 import { type ConfigContext, type ExpoConfig } from 'expo/config'
 
+const projectId = process.env.EXPO_PROJECT_ID
```
Psycarlo, line 3: Since we are grabbing projectId from env vars, we should add this env var to a `.env.example` file

Psycarlo (changes requested):

Psycarlo (approved):

PR #361, opened 2026-04-27
Refactor PIN auth + Add pin auth in missing places

v4v2:
This pull request:

## Refactors duplicate pin auth logic

Currently the logic is duplicated in:

- [x] unlock screen
- [x] view single-sig mnemonic
- [x] view multi-sig mnemonic

## Adds pin auth in missing places

- [x] ~view private descriptor single-sig~ (we do not export private descriptors yet, only public)
- [x] ~view private descriptor multi-sig~ (we do not export private descriptors yet, only public)
- [x] confirm wallet deletion

## Adds missing duress pin logic

The duress pin logic is only present in unlock screen. All other places there is missing duress pin logic.

- [x] implement shared duress pin logic


Psycarlo:
@v4v2 fix conflicts please

Psycarlo (approved):

Issue #369, opened 2026-04-29
Improvement: Sanity check on wallet creation


luca0x46:
Right now it is possible to create multiple wallets with the same name. A possible "There already exists a wallet with this name" Warning could be useful (or blocking same-name wallets altogether).

v4v2:
good catch! I will add the warning

Psycarlo:
Closed by https://github.com/satsigner/satsigner/pull/374

PR #374, opened 2026-05-06
Prevent creating account with duplicate

v4v2:
This pull request:

- [x] prevent creating wallets with same names
- [x] prevent editing the name of an existing wallet to be the name of another existing wallet
- [x] allow duplicate names for wallets of distinct networks (i.e, two wallets name "My Wallet", one is mainnet and another is testnet)
- [x] show feedback warning message to user upon all scenarios describe above
- [x] test it works on both single sig and mult-sig wallets

Solves #369.

### DIFF
--- a/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/add/index.tsx
+++ b/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/add/index.tsx
@@ -6,12 +6,14 @@ import { useShallow } from 'zustand/react/shallow'
 import SSButton from '@/components/SSButton'
 import SSCheckbox from '@/components/SSCheckbox'
 import SSText from '@/components/SSText'
-import SSTextInput from '@/components/SSTextInput'
+import SSTextInput, { type SSTextInputProps } from '@/components/SSTextInput'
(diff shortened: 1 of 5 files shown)


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/add/index.tsx
```diff
@@ -20,9 +22,38 @@ export default function Add() {
...
+    (account) => account.network === network
+  )
+
   const [localName, setLocalName] = useState('')
   const [localPolicyType, setLocalPolicyType] =
     useState<NonNullable<Account['policyType']>>('singlesig')
+  const [isValidName, setIsValidName] = useState<SSTextInputProps['status']>()
+  const [isPseudoDuplicatedName, setIsPseudoDuplicatedName] = useState(false) // pseudo-duplicate name = wallet of other network has same name
+
+  function validateName(name: string) {
```
Psycarlo: duplicate validateName function

Psycarlo: Can we make a hook?

Psycarlo (changes requested):

Psycarlo (approved):

PR #375, opened 2026-05-07
refactor

v4v2:
- [x] delete unused files
- [x] refactor some components (reorder code)
- [x] place all ark,lightning,nostr types in `@/types/models`
- [x] place all ark,lightning,nostr constants in `@/constants`
- [x] place all ark,lightning,nostr apis in their own files

This already helped removed a lot of duplicate code: constants, type declarations, utils.

For now, that is it... the PR is already big, so it is better to do further refactoring


apps/mobile/components/SSButton.tsx
```diff
@@ -309,6 +256,59 @@ function SSButton({
   )
 }
 
+function getButtonVariantStyle(
```
Psycarlo, line 259: Functions now below the component?

apps/mobile/components/SSNostrFeedTabs.tsx
```diff
@@ -76,6 +76,7 @@ type NoteKindFilterOption = {
   labelKey: string
 }
 
+// TODO: move this to constants/nostr
```
Psycarlo: Why not move now?

apps/mobile/constants/ark.ts
```diff
@@ -0,0 +1,62 @@
+import type { ArkSendKind, ArkServer } from '@/types/models/Ark'
+import type { Network } from '@/types/settings/blockchain'
+
+export function getArkServer(
```
Psycarlo: A function is not a constant. We should fix this.

Psycarlo (changes requested):

apps/mobile/components/SSButton.tsx
v4v2: yes, this is better, it is more useful this way: when opening the component file, you want to see the contents of the component, and not the helper functions. Keep what is more important at the top.

Psycarlo: @v4v2 we should add a rule for this.

Psycarlo (approved):

PR #378, opened 2026-05-09
Add zod

v4v2:
- [x] add zod
- [x] update dependency list
- [x] use Zod in `@/types/models`
- [x] use Zod in Esplora api
- [x] use Zod in Electrum client
- [x] use Zod in Blockchain api
- [x] test it works


apps/mobile/utils/bip329.ts
```diff
@@ -1,53 +1,30 @@
+import { Bip329FileType, Label } from '@/types/bips/329'
```
Psycarlo: No type keyword?

Psycarlo (changes requested):

apps/mobile/api/blockchain.ts
```diff
@@ -48,27 +49,28 @@ export class MempoolOracle implements BlockchainOracle {
     )
   }
 
-  async getAddressUtxos(address: string): Promise<UTXO[]> {
-    const data: UTXO[] = (await this.get(`/address/${address}/utxo`)) as UTXO[]
-    return data
+  async getAddressUtxos(address: string) {
+    const data = await this.get(`/address/${address}/utxo`)
+    return z.array(UTXOSchema).parse(data)
```
Psycarlo, line 54: If parse fails because of schema change or something, app will crash for all users.
Do we have a plan for this?

apps/mobile/api/electrum.ts
```diff
@@ -133,25 +69,31 @@ class BaseElectrumClient {
     const protocol = url.replace(/:\/\/.*/, '')
     const host = url.replace(`${protocol}://`, '').replace(`:${port}`, '')
 
-    // Validate host: allow domain names (including .onion), IP addresses
+    // url validation
```
Psycarlo: This comment really needed?

apps/mobile/api/electrum.ts
```diff
@@ -133,25 +69,31 @@ class BaseElectrumClient {
...
+        isValidIP,
+        isValidPort,
+        isValidProtocol,
+        url
+      })
+      throw new Error(`Backend URL ${url} seems invalid.`)
     }
 
-    const client = new ElectrumClient({
+    // additional type-safe validation with Zod
```
Psycarlo: Comment really needed?

apps/mobile/api/electrum.ts
```diff
@@ -133,25 +69,31 @@ class BaseElectrumClient {
...
-
-    if (
-      !(isValidDomain || isValidIP) ||
-      !port.match(/^[0-9]+$/) ||
-      (protocol !== 'ssl' && protocol !== 'tls' && protocol !== 'tcp')
-    ) {
-      throw new Error('Invalid backend URL')
+    const isValidPort = port.match(/^[0-9]+$/) !== null
+    const isValidProtocol = ['ssl', 'tls', 'tcp'].includes(protocol)
+    if (!(isValidDomain || isValidIP) || !isValidPort || !isValidProtocol) {
```
Psycarlo, line 76: Can't zod do this as well?

apps/mobile/api/electrum.ts
```diff
@@ -317,45 +252,49 @@ class ElectrumClient extends BaseElectrumClient {
...
-    const balance = await this.getAddressBalance(address)
     const transactions = this.parseAddressTransactions(
       address,
       rawTransactions,
       txHeights,
       txTimestamps
     )
 
-    return { balance, transactions, utxos }
+    // balance info
```
Psycarlo: Remove useless comments

apps/mobile/api/electrum.ts
```diff
@@ -317,45 +252,49 @@ class ElectrumClient extends BaseElectrumClient {
       addressKeychain
     )
 
+    // tx info
```
Psycarlo: not needed

Psycarlo (changes requested):

apps/mobile/api/esplora.ts
```diff
@@ -165,6 +193,7 @@ export default class Esplora {
     return await this._call('/fee-estimates')
   }
 
+  // test connection
```
Psycarlo: Remove

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/block/[block]/transactions.tsx
```diff
@@ -71,15 +77,15 @@ export default function BlockTransactions() {
...
-              amount: txInfo.vout.reduce((val, out) => val + out.value, 0),
-              verbosity: 1
-            }
-          })
-        )
+        const tx: BlockTx = {
+          ...txInfo,
+          amount: txInfo.vout.reduce((val, out) => val + out.value, 0),
+          verbosity: 1
+        } as BlockTx
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/block/[block]/transactions.tsx:84]: Avoid casting

Psycarlo (changes requested):

apps/mobile/api/electrum.ts
v4v2: the protocol validation can be done through enum, but the whole URL not because we use regex to check domain and IP

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/block/[block]/transactions.tsx
Comment (v4v2) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/block/[block]/transactions.tsx:84]: this is needed here. Removing casting introduces too much trouble related to typescript type annoyance

apps/mobile/api/blockchain.ts
v4v2: @Psycarlo the page component which is calling the hooks should be responsible for handling any error and not the api/*.ts file.
Moreover, the schema will only fail parsing  if the API changes its response format, and we don't expect abrupt changes in the standard API returned by Mempool instances. For this to fail, the API has to be unstable.

apps/mobile/api/electrum.ts
v4v2: the code using zod was way more verbose. To avoid re-declaring the protocol items, I'd do:

```
const protocol = ElectrumClientSchema.shape.props.shape.protocol.parse(url.replace(/:\/\/.*/, ""))
```

or

```
z.enum("tls", "tcp", "ssl").parse(url.replace(/:\/\/.*/, ""))
```

which is the same verbosity as `['ssl', 'tls' 'tcp'].includes(protocol)`

Psycarlo (approved):

PR #388, opened 2026-05-18
Fixes and Improvements

Psycarlo:
- **Removed shared-value write during render in `SSBubble`** — was mutating a Reanimated shared value inline in the render path; now done in an effect/event. Avoids a render-phase side effect.
- **Hid label placeholder in bubble when UTXO has no label** — empty-label UTXOs no longer show stray placeholder text in the bubble view.
- **Stopped reading shared value inside `useAnimationEnd` `useCallback` deps** — reading `.value` in a deps array doesn't trigger updates and breaks under react-compiler. Now reads safely.
- **Respected safe-area insets on UTXO select screens** — `selectUtxoBubbles` and `selectUtxoList` honor notch/home-indicator padding.
- **Bitcoin icon switched from filled to stroke** — visual change to `SSIconBitcoin`.
- **Fixed stale selected-UTXO total under react-compiler memoization** — total was caching stale selection; recomputes correctly now across `ioPreview`, `selectUtxoBubbles`, `selectUtxoList`.
- **Respected safe-area insets on sign-and-send screens** — same inset fix applied to `ioPreview`.
- **Stopped descriptor-validity cache from poisoning unrelated inputs** — cache key was too broad, so one input's validity leaked into others. Cache scoped properly in `SSTextInput` and `validation.ts`; also touches the unified import screen.

### DIFF
--- a/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/[id]/signAndSend/ioPreview.tsx
+++ b/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/bitcoin/account/[id]/signAndSend/ioPreview.tsx
@@ -196,7 +196,7 @@ export default function IOPreview() {
     () => utxosValue(account.utxos),
     [account.utxos]
   )
-  const utxosSelectedValue = utxosValue(getInputs())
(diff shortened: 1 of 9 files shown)


v4v2 (approved):
looks good

PR #390, opened 2026-05-21
Explorer phase 2

pedromvpg:
## Summary
- Make the explorer **node-first**: prefer configured backend (Electrum / Esplora / RPC), with explicit opt-in loads from mempool.space when the backend cannot provide a capability
- Add shared explorer capability UI/helpers, richer address/transaction/block flows, and RPC-backed chain data where needed
- Split **address**, **transaction**, and **block** into search + details routes with famous mainnet examples, loaders, and clearer detail pages
- Tag special outputs (OP_RETURN, witness commitment, empty scripts) in charts/sankey/lists; add block analytics viz and polish loaders / mempool oracle URL handling

## What's included
- **Capability layer** — shared banners/helpers so missing features explain *why* and offer a mempool.space fallback
- **Address explorer** — search + paste/scan, example addresses, details with balances/UTXOs/recent txs
- **Transaction explorer** — sample txs, decode/broadcast paths, enrichment, special-output tags, load-anyway for large hex
- **Block explorer** — height selector + famous blocks on search; details at `/explorer/block/{height}` with prev/next and analytics
- **Chain pages** — chain tip, difficulty, halving, mempool, node, network (TanStack Query + Skia viz where relevant)

## Test plan
- [ ] On mainnet with Electrum: open Address / Transaction / Block search, load an example, confirm details + loader
- [ ] Confirm capability banners appear when backend lacks a feature; "load from mempool.space" succeeds
- [ ] Transaction: open a sample with OP_RETURN / coinbase; verify tags and chart/sankey layout
- [ ] Block: open Genesis / Pizza Day / a halving; prev/next adjacent heights; view transactions when backend allows
- [ ] From chain tip / difficulty / halving, tap through to block details (`/explorer/block/{height}`)
- [ ] Repeat smoke checks on Esplora and RPC backends if available
- [ ] `pnpm --filter @satsigner/mobile typecheck` and `pnpm run check` pass


Psycarlo:
## MEDIUM

### 1. All halving milestone dots collapse to one point
**File:** `apps/mobile/components/SSHalvingProgress.tsx:141`

```js
const ratio = h.epoch % 1 === 0 && h.epoch > 0 ? 0 : 0   // both branches = 0
```

Dead ternary — returns `0` in both branches. Every past-halving dot gets `angle = START_ANGLE`, so all dots stack at `(cx, cy − TRACK_RADIUS)`. The per-halving position is never computed.

**Fix:** derive `ratio` from the halving's position on the ring, or remove the loop if dots aren't meant to be spread.

### 2. Bogus size + reversed hashes on Electrum block path
**File:** `apps/mobile/api/explorerBlock.ts:39`

`size: block.weight() * 4`, but `getBlock()` returns a **header-only** `bitcoinjs.Block`, so `weight()` is meaningless (and inverted — size ≈ weight/4, not ×4). Additionally `merkle_root` and `previousblockhash` use raw `toString('hex')` = internal little-endian order, so they render **reversed** vs canonical display order (the PR's own `blockDecoded.ts` reverses them).

**Fix:** leave `size`/`weight` undefined for the header-only Electrum path (type already allows it); reverse `merkle_root`/`previousblockhash` bytes for display.

### 3. Unconfirmed tx heights corrupt timestamps / break address load
**File:** `apps/mobile/api/explorerAddress.ts:245`

```js
const heights = limitedTxids.map((txid) => heightByTxid.get(txid) ?? 0)
```

Electrum returns height `0` for unconfirmed txs and `-1` for txs with unconfirmed parents. Height `0` fetches the **genesis** header → every unconfirmed tx gets stamped Jan-2009. Height `-1` calls `blockchainBlock_header(-1)`, which the server rejects, rejecting the whole `Promise.all` and failing the entire details load (hook has `retry: false`).

**Fix:** skip timestamp lookup for `height <= 0`; treat those txs as unconfirmed (undefined timestamp).

### 4. Nested VirtualizedLists disable virtualization
**File:** `apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/address/[address]/index.tsx` (~L142, L313, L322)

The screen body is a vertical `<ScrollView>` containing a UTXO `<FlashList>` **and** `<SSExplorerAddressTransactions>` (which renders another `<FlashList>` of txids). Same-orientation VirtualizedLists nested in a ScrollView triggers RN's "VirtualizedLists should never be nested" warning and disables virtualization — every row renders eagerly. Heavy jank/memory on high-activity example addresses.

**Fix:** render the body as a single `FlashList` with `ListHeaderComponent`, or set `scrollEnabled={false}` on inner lists and don't rely on virtualization.

### 5. Camera continuous-scan stacks duplicate screens
**Files:**
- `apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/transaction/index.tsx:166`
- `apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/address/index.tsx:78`

`CameraView.onBarcodeScanned` fires many times/sec while a QR is in frame. `setScanOpen(false)` does not stop callbacks synchronously, so multiple `router.push('/explorer/...')` calls stack duplicate detail screens.

**Fix:** guard with a one-shot `scanned` ref/flag so only the first decode is processed until the modal closes.

---

## LOW

### 6. `withTimeout` leaks its timer
**File:** `apps/mobile/hooks/useExplorerTransaction.ts:25`

`setTimeout` (30s, or 90s for `.onion`) is never cleared; stays pending after the race settles and later rejects an already-settled promise. Browsing many txs leaks timers.

**Fix:** capture the id and `clearTimeout` in a `.finally`.

### 7. `tipQuery` has no `enabled` guard
**File:** `apps/mobile/hooks/useExplorerBlock.ts:28`

Fires on every mount regardless of block state — for Electrum, opens a fresh socket each mount (even when `height` is null); for empty `backendUrl`, issues a bad fetch.

**Fix:** `enabled: Boolean(backendUrl)` (ideally only when a block is being viewed).

### 8. `sizes` field holds weight units, not bytes
**File:** `apps/mobile/api/explorerBlockTransactions.ts:28`

`sizes: sampleTxs.map((tx) => tx.weight)` — any consumer treating `sizes` as bytes/vsize is ~4× off.

**Fix:** rename to `weights`, or divide by 4 for vsize.

### 9. Historical price series is currency-case sensitive
**File:** `apps/mobile/api/blockchain.ts:231` (consumed by `useChainTipPriceHistory`)

`Reflect.get(row, currency)` — mempool.space keys prices by **uppercase** code (`USD`, `EUR`). A lowercase `currency` skips every row → silently empty series.

**Fix:** uppercase the currency before the lookup (and in the query URL).

### 10. NetworkDots divide-by-zero → NaN geometry
**File:** `apps/mobile/components/SSNetworkDots.tsx:38`

`const maxCount = Math.max(...displayEntries.map((e) => e.count))` lacks a `, 1` floor (sibling bar components have it). If distribution is non-empty but all counts are 0, `ratio = count / maxCount` is `NaN` → NaN Skia radii and touch geometry.

**Fix:** `Math.max(...displayEntries.map((e) => e.count), 1)`.

### 11. Halving center label unformatted + hardcoded English
**File:** `apps/mobile/components/SSHalvingProgress.tsx:95`

`${subsidyBtc} BTC` prints the raw float (later epochs → `0.048828125 BTC`). `Epoch ${epoch}` and `blocks to go` are hardcoded strings, not i18n.

**Fix:** format subsidy with fixed precision; route static strings through `t(...)`.

### 12. `formatCompactFiat` mis-rounds the k range (and test locks it in)
**File:** `apps/mobile/utils/priceChart.ts:66`

`formatNumber(value / 1_000, 0)` renders `1_500` as `"2k"` (1.5 → 2), `1_900` → `"2k"`. `priceChart.test.ts` asserts `formatCompactFiat(1_500) === '2k'`, locking in the bug.

**Fix:** use 1 decimal for the k range (or floor), and correct the test expectation.

### 13. ASM fallback drops witness-commitment payload
**File:** `apps/mobile/utils/specialOutput.ts:26`

ASM `"OP_RETURN <data>"` collapses to `[0x6a]`, so a BIP141 witness commitment supplied as an ASM string is misclassified as generic OP_RETURN. Latent — all current callers pass raw script bytes, not ASM — but `scriptToBytes` advertises ASM support.

**Fix:** parse the ASM data payload, or document that witness-commitment detection requires byte/hex input.

### 14. Scanned QR not stripped of BIP21 URI
**File:** `apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/address/index.tsx:78`

Scanned data is passed to `navigate` with only `.trim()`. Address QRs are commonly BIP21 (`bitcoin:bc1...?amount=`); `resolveAddressParam` only does `decodeURIComponent`+`trim`, so a standard address QR yields a failed lookup.

**Fix:** strip a `bitcoin:` scheme and query params before navigating.

### 15. TxID auto-nav checks length only, not hex
**File:** `apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/transaction/index.tsx:130`

`handleInputContent` auto-navigates when `trimmed.length === TXID_LENGTH` — length only, unlike `detectInputType` (requires `HEX_REGEX`). A 64-char non-hex paste/scan navigates to a screen that can only error.

**Fix:** reuse `detectInputType(...) === 'txid'` before auto-navigating.

### 16. Difficulty chevron bounds keyed to stale epoch
**File:** `apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/difficulty.tsx`

`atMinEpoch`/`atMaxEpoch` derive from `Number(fetchedEpoch ?? inputEpoch)`, but `nextEpoch`/`previousEpoch` step off `Number(inputEpoch)`. After a fetch, typing a new value leaves chevron enabled/disabled state keyed to stale `fetchedEpoch` → bounds mismatch visible input. Cosmetic, no crash.

### 17. Ancestor-linking loop is worst-case O(n³)
**File:** `apps/mobile/hooks/useInputTransactions.ts:729`

The `while (expanded)` loop rebuilds `Array.from(filteredTransactions.values())` inside a nested scan. Fine for small sankey sets; cache the array / precompute a spent-outpoint `Set` if inputs can grow large.

Psycarlo:
## 1. Sankey flow shows zero fee on electrum / rpc / hex-fallback

`app/(authenticated)/(tabs)/(signer,explorer,converter)/explorer/transaction/[txid]/index.tsx:120-131`

```ts
const totalIn = nonCoinbaseInputs.reduce((sum, inp) => sum + (inp.value ?? 0), 0)
const minerFeeSats = totalIn > 0 ? Math.max(0, totalIn - totalOut) : 0
```

Only `fetchFromEsplora` populates `inp.value` (from `prevout`). `explorerTxFromHex` — used by electrum, rpc, and the mempool hex fallback — never sets it. So `totalIn === 0` → `minerFeeSats = 0`.

Downstream in `hooks/useNodesAndLinks.ts`:
- `minerFeeSats !== undefined` forces `remainingBalance = 0`
- `minerFee > 0` is false → miner-fee node is **not pushed at all**

Result: on electrum/rpc the flow diagram renders a transaction with no fee arm.

`buildInputsMap` compounds it — it back-solves missing input values as `(totalOutputValue - knownTotal) / missingCount`. With `knownTotal = 0` each input gets `totalOut / n`, so inputs sum exactly to outputs, cementing fee = 0.

**Fix:** pass `minerFeeSats: undefined` when no real prevout values exist, so the chart falls back to `feeRate × vsize` instead of asserting an exact fee of zero.

---

## 2. `useExplorerAddress` fires a query that always rejects on RPC backend

`hooks/useExplorerAddress.ts:31-33`

```ts
const query = useQuery({
  enabled: ready,   // ignores backendSupported
```

`capability` / `backendSupported` are computed but never used to gate the query. With `backend === 'rpc'`, `fetchExplorerAddressFromBackend` returns `Promise.reject(new Error('rpc_unsupported'))` (`api/explorerAddress.ts:172`). Every RPC address view fires the query, fails, and React Query retries (default 3×) before the capability banner is reachable.

`hooks/useExplorerBlockRawHex.ts:26` gets this right:

```ts
const enabled = ready && (capability.available || mempoolEnabled)
```

**Fix:** apply the same gate in `useExplorerAddress`.

---

## 3. Electrum address tx details: redundant fetch + silent height loss

`api/explorerAddress.ts:236-247`

```ts
const limitedTxids = txids.slice(0, ESPLORA_ADDRESS_TXS_PER_REQUEST)
...
const history = await client.getAddressTransactions(address)
const heightByTxid = new Map(history.map((e) => [e.tx_hash, e.height]))
const heights = limitedTxids.map((txid) => heightByTxid.get(txid) ?? 0)
```

Two problems:

1. `getAddressTransactions` re-fetches the full history that `fromElectrum` already fetched — wasted round-trip on every detail load.
2. Any txid in `limitedTxids` missing from `history` silently falls back to `height = 0`. `getBlockTimestamps` now maps `height <= 0 → undefined` (`api/electrum.ts:308-313`), so those txs render with **no timestamp** rather than surfacing an error.

**Fix:** pass the already-known heights from the caller instead of refetching.

---

## 4. Bitnodes node lookup hardcodes mainnet port 8333

`api/explorerNode.ts:151`

```ts
const nodeRes = await fetch(`${BITNODES_API}/nodes/${host}-8333/`)
```

The lookup key is hardcoded to the mainnet port regardless of `selectedNetwork`, and Bitnodes indexes mainnet only. On testnet/signet this queries a **mainnet** node record for the same hostname — returning either 404 (handled) or, worse, a real mainnet node's height and user-agent presented as the user's testnet node.

Also dead code: the preceding `/snapshots/?limit=1` fetch is checked for existence, then its result is never used.

**Fix:** derive the port from the configured server URL, and skip the lookup entirely when the network is not mainnet.

---

## 5. `fetchBitnodesNetworkStats` has no error handling

`api/explorerNode.ts:180-215`

Unlike `fetchBitnodesNodeInfo`, this function has no `try`/`catch` and no `res.ok` check on either fetch:

- A non-200 from bitnodes makes `.json()` throw on the HTML error body.
- `snapshot.results?.[0]` is guarded, but `nodesData.nodes` is not — a malformed payload throws inside `Object.values(nodesData.nodes)`.

The error propagates raw to the wrapping query.

**Fix:** check `res.ok` on both fetches and guard `nodesData.nodes` before iterating.

pedromvpg:
Thanks — all five addressed in `fab3e4709`.

**1. Zero fee on electrum / rpc / hex-fallback flow** — `explorerHistoryFee` now detects when no non-coinbase input has a known prevout value and returns `minerFeeSats: undefined` instead of an exact `0`. `useNodesAndLinks` already treats `undefined` as "fall back to feeRate × vsize" and stops forcing `remainingBalance = 0`, so the chart no longer asserts a fee of zero when it's genuinely unknowable.

**2. `useExplorerAddress` fires a query that always rejects on RPC** — applied the same gate `useExplorerBlockRawHex` uses: `enabled = ready && (backendSupported || mempoolEnabled)`. RPC address views no longer fire → fail → retry.

**3. Electrum address tx details: redundant fetch + silent height loss** — `fromElectrum` now returns a `heightByTxid` map (heights it already fetched), threaded through the page → component → hook → `fetchExplorerAddressTxDetailsFromBackend`. The details fetch uses those known heights and only re-fetches history when they're absent, removing both the extra round-trip and the silent `height = 0`.

**4. Bitnodes node lookup hardcodes mainnet port 8333** — `fetchBitnodesNodeInfo` now takes `network`, returns `null` on testnet/signet (Bitnodes indexes mainnet only), and derives the port from the configured server URL (default 8333). Dead `/snapshots/?limit=1` fetch removed; `useBitnodesNodeInfo` also gates `enabled` to mainnet.

**5. `fetchBitnodesNetworkStats` has no error handling** — wrapped in try/catch returning empty stats, added `res.ok` checks on both fetches, and guarded `nodesData.nodes` before iterating.

Added unit tests for the height threading and the Bitnodes node / network-stats paths. `pnpm run check`, `typecheck`, and `test:unit` (1545 tests) all pass.

Psycarlo (approved):

PR #391, opened 2026-05-21
Onboard

pedromvpg:


Psycarlo:
@pedromvpg status on this one?
A LOT of conflicts

Psycarlo:
## 🔴 Blocker

### 1. Intro overlay shows on every cold start, covering the unlock screen

`apps/mobile/store/intro.ts:17` — the store defaults to `visible: true`, is not persisted, and is not gated on `firstTime`. On **every** cold launch, a returning PIN-protected user gets a non-skippable, touch-blocking 1.4s animation (`zIndex: 9999`, opaque background) rendered on top of `/unlock` before they can authenticate.

The only intentional trigger in the PR is `showIntro(true)` from Settings → About — nothing ever deliberately invokes the returning-user path; it fires purely from the un-gated default.

**Fix:** gate the overlay on `firstTime` (or persist a `hasSeenIntro` flag), so it renders only on first run and on explicit replay from About.

## 🟠 Should fix

### 2. ~21 wasted animated-style worklets per frame on the coordination step

`apps/mobile/components/SSIntroAnimationCoordinationStep.tsx:342` — `CoordinationNode` builds 8 `useAnimatedStyle` hooks unconditionally, but phone nodes never attach `indicator0/1/2Style`/`checkStyle` and the circle node never attaches `borderStyle`. Verified against Reanimated source: unattached style worklets still execute on the UI thread every frame. With the infinite `cycleScript` loop, that's ~21 discarded worklet evaluations per frame at 60fps for as long as the step is mounted (no auto-advance).

**Fix:** split into `PhoneNode`/`DescriptorNode` so each only builds the styles it attaches. This also collapses the three copy-pasted `indicatorNStyle` blocks into one parameterized helper.

### 3. No tests for the new feature

The PR adds a new store plus 9 components (~3000 lines) with zero tests. CLAUDE.md expects tests for new features, and every comparable store has one in `tests/unit/store/`. A simple `intro.test.ts` covering show/hide/`forceFirstTime` and the first-time-vs-returning gating would have caught issue #1.

## 🟡 Decide consciously

### 4. `InteractionManager` → `requestIdleCallback` changes semantics

`apps/mobile/hooks/useNostrMessageProcessor.ts:101` (also `SSTransactionChart.tsx:158`, `transaction/[txid]/index.tsx:142`) — no crash (RN 0.86 provides the global), but RN's `requestIdleCallback` is frame-tick driven: it pauses while the app is backgrounded (Nostr batch processing can stall until foreground) and has no awareness of UI-thread animations, so the "wait for animations to finish" guarantee the deleted comments described no longer holds. Fine to keep since `InteractionManager` is deprecated — but it should be a conscious choice, ideally noted in the code.

Psycarlo (approved):

PR #405, opened 2026-06-09
feat: converter written numbers, European scale, date-anchored price conversion

pedromvpg:
image

image

## Summary

- **Written numbers toggle** — shows amounts spelled out (e.g. "~30 billion sats") below each currency value, with smooth animated height expand/collapse on toggle
- **European scale toggle** — switches written numbers between short scale (billion/trillion) and long scale (milliard/billiard/trillion), enabled only when written numbers are on
- **Date-anchored conversion** — changing the date picker now re-converts from the last-edited field using the new date's prices (e.g. "how many sats was 1000 USD in 2014?")
- **Date picker UX** — day column now shows weekday abbreviation alongside the day number
- **Currency input fixes** — `includeFontPadding={false}` and `textAlignVertical="center"` for correct vertical alignment inside inputs

## New files

- `apps/mobile/utils/formatLargeNumber.ts` — pure utility supporting both short scale and long (European) scale number naming

## Test plan

- [ ] Toggle "Written Numbers" on/off — values animate in/out with no layout shift on other elements
- [ ] Toggle "European Scale" — labels switch between billion/milliard etc.
- [ ] Enter a fiat value (e.g. 1000 USD), change date — USD stays fixed, sats/BTC update to reflect that date's price
- [ ] Enter a sats/BTC value, change date — sats/BTC stays fixed, fiat values update
- [ ] Press "Today" button — resets date and re-fetches current prices
- [ ] Date picker day column shows weekday abbreviation

🤖 Generated with [Claude Code](https://claude.com/claude-code)


apps/mobile/utils/formatLargeNumber.ts
```diff
@@ -0,0 +1,37 @@
+const SHORT_SCALE = [
+  { value: 1e15, word: 'quadrillion' },
+  { value: 1e12, word: 'trillion' },
```
Psycarlo, line 3: These words are not i18n'd.

Also, can you use `Intl` API? Won't work with `quadrillion`, but you can handle that case.

Psycarlo (changes requested):

pedromvpg:
Closing in favor of #406, which already includes this converter work plus the action-group gradient style. Review feedback from this PR is addressed there.

PR #406, opened 2026-06-09
feat: converter written numbers + action group gradient border

pedromvpg:
image

image

### Converter
- **Written numbers toggle** — spells out amounts below each field (e.g. `~30 billion sats`) with animated expand/collapse
- **European scale toggle** — switches between short scale (billion/trillion) and long scale (milliard/billiard/trillion); only enabled when written numbers are on
- **Date-anchored conversion** — changing the date re-converts from the last-edited field using that date's prices (e.g. "how many sats was 1000 USD in 2014?")
- **Date picker UX** — weekday abbreviation in the day column; month/day names i18n'd under `date.*`
- **Currency input alignment** — improved vertical alignment via `textAlignVertical`

### Action group style
- Replaces uniform `borderTopColor` with a 1px `LinearGradient` strip per button
- Simulates a single overhead light source: center brightest, outer edges fade darker
- Applies to both compact (3-button) and full (5-button) layouts

### Review feedback addressed
- `formatLargeNumber` moved to `utils/format` and uses `Intl.NumberFormat` with i18n for quadrillion / long-scale edge cases
- Removed redundant one-line currency handlers; call `handleValueChange` directly
- Replaced nested `useFocusEffect`/`useCallback` with `useMountEffect`
- Added `isToday` helper in `utils/date`
- Unit tests for `formatLargeNumber` and `isToday`

Supersedes #405.

## Test plan

- [ ] Toggle Written Numbers on/off — values animate without layout shift
- [ ] Toggle European Scale — labels switch between billion/milliard etc.
- [ ] Enter fiat (e.g. 1000 USD), change date — USD stays fixed, sats/BTC update
- [ ] Enter sats/BTC, change date — sats/BTC stay fixed, fiat values update
- [ ] Today button disables when date is already today
- [ ] Date picker shows weekday abbreviations and localized month names
- [ ] Action group buttons show gradient top border (compact + full layouts)


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/converter/currency.tsx
```diff
@@ -72,185 +196,416 @@ export default function Converter() {
...
 
-  useFocusEffect(() => {
-    const timestamp = Math.floor(date.setHours(0, 0, 0, 0) / 1000)
+  function handleSatsChange(value: number) {
+    handleValueChange('sats', value)
+  }
+  function handleBitcoinChange(value: number) {
+    handleValueChange('bitcoin', value)
+  }
+  function handleUsdChange(value: number) {
```
v4v2: instead of multiple redundant one-line functions, remove them and call the `handleValueChange` directly, and whenever it is called, pass both currency and value as arguments. Lets simplify things.

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/converter/currency.tsx
```diff
@@ -72,185 +196,416 @@ export default function Converter() {
...
+  )
+
+  useFocusEffect(
+    useCallback(() => {
+      if (!hasInitializedRef.current) {
+        hasInitializedRef.current = true
+        handleValueChange('bitcoin', 1)
+      }
+    }, []) // eslint-disable-line react-hooks/exhaustive-deps
+  )
```
v4v2, line 268: do not nest hooks such as `useFocusEffect`, `useCallback`, `useEffect`. If `useEffect` is not enough, just use `useEffect`. Also, since we have compiler to optimize build, all uses of `useCallback` are unnecessary.

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/converter/currency.tsx
```diff
@@ -72,185 +196,416 @@ export default function Converter() {
...
+                markHeight={DATE_PICKER_MARK_HEIGHT}
+                startYear={DATE_PICKER_START_YEAR}
               />
+              <SSVStack gap="sm" style={styles.toggleRow}>
+                <SSButton
+                  key="today"
+                  label={t('date.today')}
+                  variant="outline"
+                  onPress={handleTodayPress}
+                  disabled={date.toDateString() === new Date().toDateString()}
```
v4v2: extract this to the beginning of the component, via helper function

```
function isToday(date) {
  return date.toDateString() === new Date().toDateString()
}
```

then

```
disabled={isToday(date)}
```

apps/mobile/components/SSDatePicker.tsx
```diff
@@ -120,6 +130,44 @@ function SSDatePicker({
...
+    'May',
+    'June',
+    'July',
+    'August',
+    'September',
+    'October',
+    'November',
+    'December'
+  ]
+  const DAY_NAMES = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat']
```
v4v2: move this to `locales.json` under a new `date` section (root section).

```
"date": {
  "jan": "january",
  "feb": "february,
  ...
  "sun": "sunday",
}
```

So we can handle other languages later on.

Ensure:

- flat structure of `locale.json`, we allow only 1 level of nesting.
- short key names for the dates (prefer `jan`, `feb, `oct`, since the first three letters of all month names and all day names are distinct, we can make it shorter, less verbose)

apps/mobile/utils/formatLargeNumber.ts
```diff
@@ -0,0 +1,37 @@
+const SHORT_SCALE = [
```
v4v2: move the large-number words to `locales.json`, maybe under `numbers` section (maybe AI can suggest more suitable section name). Apply the suggestion from the previous requested change.

apps/mobile/utils/formatLargeNumber.ts
```diff
@@ -0,0 +1,37 @@
...
+  { value: 1e21, word: 'trilliard' },
+  { value: 1e18, word: 'trillion' },
+  { value: 1e15, word: 'billiard' },
+  { value: 1e12, word: 'billion' },
+  { value: 1e9, word: 'milliard' },
+  { value: 1e6, word: 'million' },
+  { value: 1e3, word: 'thousand' }
+]
+
+export function formatLargeNumber(num: number, european = false): string {
```
v4v2: we have `utils/format`, place it there. Delete this file.

v4v2 (dismissed):

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/converter/currency.tsx
Psycarlo: @pedromvpg  We should avoid using `useEffect` when possible.
If truly needed, like to run things on mount, use the `useMountEffect` hook instead.

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/converter/currency.tsx
Psycarlo: I would even extract `isToday` function to utils and add unit tests.

apps/mobile/utils/formatLargeNumber.ts
Psycarlo: What about just using `Intl.NumberFormat`?

```
export function formatLargeNumber(num: number, locale = 'en'): string {
  if (!isFinite(num) || num === 0) return ''

  const opts: Intl.NumberFormatOptions = {
    notation: 'compact',
    compactDisplay: 'long'
  }
  const exact = new Intl.NumberFormat(locale, { ...opts, maximumFractionDigits: 20 }).format(num)
  const rounded = new Intl.NumberFormat(locale, { ...opts, maximumFractionDigits: 0 }).format(num)

  return (exact !== rounded ? '~' : '') + rounded
}
```

Only tradeoff is that `Intl.NumberFormat` only works until trillion.
For `quadrillion` would need custom code, something like:

```
export function formatLargeNumber(
  num: number,
  locale = 'en',
  quadrillionLabel?: (count: number) => string
): string {
  if (!isFinite(num) || num === 0) return ''

  if (Math.abs(num) >= 1e15) {
    const r = Math.round(num / 1e15)
    const approx = r * 1e15 !== num ? '~' : ''
    const word = quadrillionLabel
      ? quadrillionLabel(Math.abs(r))
      : `quadrillion${Math.abs(r) > 1 ? 's' : ''}` // en fallback
    return `${approx}${r} ${word}`
  }

  const opts: Intl.NumberFormatOptions = { notation: 'compact', compactDisplay: 'long' }
  const exact = new Intl.NumberFormat(locale, { ...opts, maximumFractionDigits: 20 }).format(num)
  const rounded = new Intl.NumberFormat(locale, { ...opts, maximumFractionDigits: 0 }).format(num)
  return (exact !== rounded ? '~' : '') + rounded
}
```

apps/mobile/utils/formatLargeNumber.ts
Psycarlo: Agreed. Also add unit tests.

pedromvpg:
Addressed all review feedback in `0e6093f9b` and `740d018c4`:

**@v4v2**
- Removed redundant one-line currency handlers; inputs call `handleValueChange(key, value)` directly
- Replaced nested `useFocusEffect`/`useCallback` with `useMountEffect` for first-load init and a plain (non-nested) `useFocusEffect` for price refetch on screen focus
- `isToday(date)` extracted to `utils/date` and used for the Today button
- Month/day names moved to `locales/en.json` under `date.*` (`jan`, `feb`, `sun`, etc.)
- `formatLargeNumber` moved to `utils/format`; deleted `formatLargeNumber.ts`
- Large-number edge-case words under `numbers.*` in locales

**@Psycarlo**
- `formatLargeNumber` uses `Intl.NumberFormat` (`compact` + `long`) with i18n-backed handling for quadrillion / European long-scale (≥1e15)
- `useMountEffect` used for mount-only init instead of `useEffect`
- `isToday` in `utils/date` with unit tests
- Unit tests added for `formatLargeNumber` in `format.test.ts`

Ready for re-review.

Psycarlo (approved):

PR #418, opened 2026-06-19
Fix child data not deleted upon child deletion.

v4v2:
- [x] Fix child data not deleted upon child deletion.
- [x] Refactor duplicate code (update account row)


Psycarlo (approved):

Psycarlo:
We need a way to test these things, so we are confident we are deleting the data we want, no more, no less

PR #422, opened 2026-06-23
remove some duplicate code in ark pages

v4v2:
- [x] refactor duplicated styesheet
- [x] refactor duplicated UI code that displays fee estimates and total amount
- [x] place utility functions to bottom of file (tsx components are more important, thus shall come first)

### DIFF
--- a/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx
+++ b/apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx
@@ -38,6 +38,8 @@ import { millisatsToSats } from '@/utils/bitcoinUnits'
 import { formatFiatPrice, formatNumber } from '@/utils/format'
 import { validateAddress } from '@/utils/validation'
 
+import { styles } from '../../[id]/pay-invoice'
+
 function destinationDisplay(draft: ArkDestinationDraft): string {
   if (draft.kind === 'arkoor') {
     return truncateArkCounterparty(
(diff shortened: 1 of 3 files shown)


apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx
```diff
@@ -38,91 +38,7 @@ import { millisatsToSats } from '@/utils/bitcoinUnits'
...
-function feeKindFromDraft(draft: ArkDestinationDraft): ArkSendFeeKind {
-  if (draft.kind === 'arkoor') {
-    return 'arkoor'
-  }
-  if (draft.kind === 'onchain') {
-    return 'onchain'
-  }
-  return 'lightning'
-}
+import { ArkInvoiceStats, styles } from '../../[id]/pay-invoice'
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx:41]: Relative import?

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx
```diff
@@ -452,6 +336,92 @@ export default function ArkSendConfirmPage() {
   )
 }
 
+function destinationDisplay(draft: ArkDestinationDraft): string {
```
Comment (Psycarlo) [inline apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx:339]: For now, good.
But we should move them outside of the page component file eventually.

Psycarlo (changes requested):

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx
Comment (v4v2) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx:339]: true. But I didn't want to pollute the `components/` directory with a component that only shows up twice in such a narrow context

apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx
Comment (v4v2) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx:41]: I usually dont like relative imports (I am strongly against it in the context of `@/components/`, `@/hooks/`, `@/utils/`), but in the context of pages importing from other pages I think it is better this way

Comment (Psycarlo) [reply apps/mobile/app/(authenticated)/(tabs)/(signer,explorer,converter)/signer/ark/account/[id]/send/confirm.tsx:41]: Can skip for now, but we should follow 1 rule instead of 2. I would go with `@` imports.

Psycarlo (approved):

PR #423, opened 2026-06-23
fix: ensure toast notifications appear above any modal dialog

### LINKED ISSUE #395: Title: Toasts hidden behind modals
## Problem

Toast notifications triggered while modals are open are not visible. Example: dust validation `toast.error()` on "Add Output" modal in `ioPreview` fires but user sees nothing.

## Root cause

- `Toaster` from `sonner-native` mounted once in [`apps/mobile/app/_layout.tsx:132`](apps/mobile/app/_layout.tsx#L132)
- `SSModal` ([`apps/mobile/components/SSModal.tsx:32`](apps/mobile/components/SSModal.tsx#L32)) wraps content in React Native's built-in `Modal`
- RN `Modal` renders in a separate native window/view hierarchy, so the `Toaster` (sibling in the app tree) is rendered beneath it regardless of `zIndex` (toaster has `999999`, modal has `1000` — irrelevant across the native boundary)

## Repro

1. Open an account → Sign and Send → ioPreview
2. Tap **Add Output**
3. Enter an amount below dust limit and submit
4. **Expected:** toast error visible
5. **Actual:** nothing shown (toast renders behind modal)

v4v2:
changes:

- [x] ensure toast notifications appear above any modal dialog 

Code is ready, but needs testing:

- [x] manually test all base-modal components: ssmodal, ssgradientmodal, sswarningmodal, ssselectmodal

Fixes #395

### DIFF
--- a/apps/mobile/components/SSGradientModal.tsx
+++ b/apps/mobile/components/SSGradientModal.tsx
@@ -24,7 +24,7 @@ function SSGradientModal({
   return (
     <Modal visible={visible} transparent animationType="fade">
       <SSMainLayout style={styles.containerBase}>
-        <Toaster /> {/* ensures toasts render above modal */}
+        <Toaster />
         <LinearGradient
           style={styles.gradientContainerBase}
           colors={[Colors.gray[950], Colors.gray[800]]}
--- a/apps/mobile/components/SSGradientModal.tsx
+++ b/apps/mobile/components/SSGradientModal.tsx
@@ -1,5 +1,6 @@
 import { LinearGradient } from 'expo-linear-gradient'
 import { Modal, StyleSheet } from 'react-native'
+import { Toaster } from 'sonner-native'
 
 import SSMainLayout from '@/layouts/SSMainLayout'
 import { t } from '@/locales'
@@ -23,6 +24,7 @@ function SSGradientModal({
   return (
     <Modal visible={visible} transparent animationType="fade">
       <SSMainLayout style={styles.containerBase}>
+        <Toaster /> {/* ensures toasts render above modal */}
         <LinearGradient
           style={styles.gradientContainerBase}
           colors={[Colors.gray[950], Colors.gray[800]]}
--- a/apps/mobile/components/SSModal.tsx
+++ b/apps/mobile/components/SSModal.tsx
@@ -38,8 +38,8 @@ function SSModal({
           { paddingBottom: insets.bottom, paddingTop: insets.top }
         ]}
       >
+        <Toaster />
         <View style={styles.container}>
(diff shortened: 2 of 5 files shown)


apps/mobile/components/SSGradientModal.tsx
```diff
@@ -23,6 +24,7 @@ function SSGradientModal({
   return (
     <Modal visible={visible} transparent animationType="fade">
       <SSMainLayout style={styles.containerBase}>
+        <Toaster /> {/* ensures toasts render above modal */}
```
Psycarlo: Won't we have a bunch of <Toaster /> components mounted, this way?

Psycarlo (changes requested):

apps/mobile/components/SSGradientModal.tsx
v4v2: @Psycarlo

1. We will have multiple `<Toaster />` mounted only when there are multiple modals mounted. No modals -> no multiple toasters.
2. Each `<Modal>` gets its own window frame, separated from root layout, thus each `<Toaster />` lives in a distinct window frame and won't interfer with each other.
3. It's possible to mount multiple `<Toast />` components and all of the components have access to toasts emitted anywhere in the app (it's like a singleton architecture)

I just need to double check one thing: if modal is hidden, the toasts notifications inside that modal's window will be hidden too.

---

There is another solution (I asked AI) but it would require refactoring our root layout architecture and install a new library.
For now, let's go for it.

v4v2: AI's feedback:

Use a Portal-aware toast (Recommended for scale). Replace or augment sonner-native with [@gorhom/portal](https://github.com/gorhom/react-native-portal) so toasts render into a host that lives at the top of the native stack.
```typescript
tsx// apps/mobile/app/_layout.tsx
import { PortalHost, PortalProvider } from '@gorhom/portal'

export default function RootLayout() {
  return (
    <PortalProvider>
      {/* ...rest of your providers */}
      <Stack />
      <PortalHost name="toast" />   {/* ← topmost host */}
    </PortalProvider>
  )
}
```
Then wrap your Toaster in a Portal:
```typescript
tsx// apps/mobile/components/ToastPortal.tsx
import { Portal } from '@gorhom/portal'
import { Toaster } from 'sonner-native'

export function ToastPortal() {
  return (
    <Portal hostName="toast">
      <Toaster />
    </Portal>
  )
}
```

And render <ToastPortal /> inside SSModal (or globally). This ensures the toast always renders in the PortalHost at the top of the tree, above any Modal.

Psycarlo:
Make sure it works and we will merge it.

Psycarlo (approved):

PR #428, opened 2026-06-25
feat(mobile): Bitcoin Core RPC backend support

pedromvpg:
## Summary

- Adds Bitcoin Core JSON-RPC as a third network backend alongside Electrum and Esplora, including a new `BitcoinRpc` client with Android localhost remapping and actionable error messages.
- Upgrades `react-native-bdk-sdk` to 0.5.x and wires RPC into wallet sync, transaction broadcast, input fetching, connection testing, and explorer chain tip.
- Extends network settings with RPC credentials, scan-from-block height, block-filter index warnings, wallet rescan, and uppercase RPC labels in the backend list.

### Configurable fiat price data
- New **Fiat Data** settings screen to control how fiat price data is fetched and displayed.
- Independent toggles for **fetching current spot prices** and **fetching historical prices** (historical is off by default for privacy).
- **Price provider** selection: Mempool or a custom Mempool-compatible API base URL.
- Privacy disclaimer that fetching historical prices sends transaction timestamps to the provider (can reveal when you transacted).
- New `useFiatData` hook plus `constants/fiatPriceApi.ts` / `utils/fiatData.ts`; transaction cards, account cards, chain tip, history chart, and ecash/lightning views now respect the current/historical display settings.

### Transaction card & chart UI
- Reworked transaction card balance display and layout — running balance delta with percentage change, tighter/expand-aware layout, and detail/header polish.
- Styled sat amounts dim the leading zero-padding digits for readability.
- Sankey transaction chart labels outputs as **spent**, **change**, or **self-send**.

### RPC sync progress
- The progress indicator previously always read `100%` because incremental syncs skipped the block rescan and hit a hardcoded completion value. It now emits `(currentHeight, tipHeight)` (mapping Core's `scanning.progress` fraction back to a block height) and shows `Scanning block current / tip — pct%` during a real rescan; incremental syncs show `Syncing…` instead of a misleading `100%`.

### Transaction list performance
- Reuse prior `Transaction` object references for unchanged rows across syncs so only changed rows re-render.
- Give the `FlashList` a full-height viewport (removed a stray `height: 400` that capped it), raise `drawDistance`, cap the per-row reveal animation to the first screenful, and lighten each cell (2-span styled sat text instead of one span per character, single `parseLabel` call, non-live `TimeAgo` in the list).

### Smaller fixes
- **Dev backup/restore** — RPC connection data (server URL, backend, credentials, wallet name, scan-from height) round-trips via `serverSettings`; also include the per-account `birthdayDate` and `rpcLastBlockHash` in the backup so a restored RPC wallet keeps its birthday (no forced full rescan) and sync checkpoint.
- **Wallet birthday** — inline validation feedback on the birthday date input (invalid dates now surface an error instead of silently doing nothing).
- **RPC unit tests** — added `tests/unit/api/rpc.test.ts` and `tests/unit/api/bdkRpc.test.ts`.
- **Repo hygiene** — stop tracking an accidental npm `package-lock.json` (this is a pnpm monorepo) and gitignore it; gitignore the local `scripts/test-rpc-descriptor.mjs` debug script; pin `react-native-bdk-sdk` to exact `0.5.0`.
- **Lint/type cleanups** — fix a pre-existing `SSTimeAgoText` style type error, remove redundant `?? []` inside `new Set(...)`, and normalize import ordering/formatting.

## Test plan

- [ ] Add a custom Bitcoin Core RPC server (mainnet/testnet/signet) with username/password and verify connection test succeeds
- [ ] Confirm block-filter index warning appears when `blockfilterindex=1` is not enabled
- [ ] Create or unlock a singlesig account on an RPC backend and verify wallet sync completes with progress
- [ ] Verify sync progress shows `Scanning block X / Y — Z%` during a full rescan and `Syncing…` (not `100%`) on incremental syncs
- [ ] Sign and broadcast a transaction; confirm it appears via RPC/`getrawtransaction`
- [ ] Open a transaction detail screen and verify input transactions load
- [ ] Use wallet rescan from account settings and confirm a full rescan starts on next unlock
- [ ] Open Explorer chain tip with RPC backend selected and verify height, mempool, and fee data
- [ ] Confirm address-only (watch-only) accounts show the expected error when syncing on RPC
- [ ] Test on Android emulator with `localhost` / `127.0.0.1` host — should remap to `10.0.2.2`
- [ ] In Fiat Data settings, toggle current/historical price fetching and switch providers; confirm transaction cards and account balances reflect the choices and that historical prices are only fetched when enabled
- [ ] Enter an invalid wallet birthday (e.g. `2024-13-01`) and confirm an inline validation error is shown
- [ ] Open the Sankey transaction chart and confirm outputs are labeled spent / change / self-send
- [ ] Scroll the transaction list quickly on a release build and confirm rows render without persistent blank areas
- [ ] Back up from Settings > Dev, restore on a fresh install, and confirm the RPC server (url/credentials/wallet name) and each account's birthday and sync checkpoint are restored


v4v2:
@pedromvpg how is it going ? wanna help testing ? Also, it would be awesome if the RPC works for regster, because it is way easier to spin up a regtest node to test.

Psycarlo:
@pedromvpg Remove console.logs

Psycarlo:
@pedromvpg 

## Must fix before merge
- [ ] **Android URL remap not applied to BDK sync/broadcast path** — `adjustRpcUrl()` (localhost → 10.0.2.2) only exists in `api/rpc.ts` for the JS client. `syncWallet()` and `broadcastTransaction()` in `api/bdk.ts` construct `BdkRpcClient` with the raw `server.url`. On an Android emulator with `127.0.0.1`, the connection test passes but wallet sync and broadcast fail. Export `adjustRpcUrl` and apply it before constructing `BdkRpcClient`.

- [ ] **`adjustRpcUrl` remaps any `172.*` host** — `hostname.startsWith('172.')` matches the whole `172.0.0.0/8` range, so a physical device pointed at a LAN node on `172.16.x.x` gets silently rewritten to `10.0.2.2` and can never connect. Also runs on all Android devices, not just emulators. Drop the `172.` branch (keep only `localhost`/`127.0.0.1`), ideally gated on `isEmulator()` from `react-native-device-info` (already a dependency).

- [ ] **Persisted `connectionMode` flip can permanently disable auto-polling** — `settings/network/[network].tsx` writes `connectionMode: 'manual'` into the MMKV-persisted blockchain store on mount and restores it in unmount cleanup. If the app is killed while the screen is open, the user's `auto` mode is silently lost forever. Use an in-memory suppression flag instead of mutating persisted config.

- [ ] **`rpcWalletName` collected but never used** — the form field, `Server.rpcWalletName` schema, and round-tripping exist, but no RPC call appends `/wallet/<name>` to the URL. Users with multi-wallet nodes will set it and silently get the default wallet. Wire it into the RPC URL path in both `BitcoinRpc` and the `BdkRpcClient` construction, or remove the field.

- [ ] **Empty port silently connects to port 80** — `constructUrl()` in `useCustomNetworkForm.ts` produces `http://host` when the port is empty, while the placeholder implies the network default (8332/38332/18332) is used. Apply `defaultRpcPort(networkType)` when empty, or make the port required for RPC.

- [ ] **Always-`undefined` `blockHeight`** — `hooks/useInputTransactions.ts`: `blockHeight: rawTx.confirmations ? undefined : undefined` — both branches identical. Either compute it (`currentTip - confirmations + 1`) or write plain `blockHeight: undefined`; current code reads like unfinished work.

Psycarlo (approved):

PR #441, opened 2026-07-14
fix distinct height between label input and tag input

### LINKED ISSUE #425: Tag input and label input have different height
### Context

Tag input and label input have different height.
Also, tag input has placeholder and label input does not.
Looks weird.

### Solution

Increase tag input height to match label input height.
Remove tag input placeholder text.

### Image

Image

v4v2:
This affects in all label screens: add/edit tx/utxo/address label.

Fixes #425.


Psycarlo:
@v4v2 What about "Also, tag input has placeholder and label input does not."?

apps/mobile/components/SSTagInput.tsx
```diff
@@ -74,12 +74,16 @@ function SSTagInput({
...
             blurOnSubmit={false}
-            placeholder={t('common.addTag')}
             align="left"
             size="small"
             ref={(ref: TextInput | null) => {
               inputRef.current = ref
             }}
+            style={{
+              height: 'auto',
+              padding: 10,
```
Psycarlo: This is the best way to fix it? Or should we match height another way?

Psycarlo (changes requested):

Psycarlo (approved):

Psycarlo:
With my changes, It now allows 1 char tags.

PR #449, opened 2026-07-17
feat(mobile): wallet ownership, Sankey charts, labels backup, and Nostr device seeds

pedromvpg:
## Summary
- Annotate synced and optimistic transactions with Sparrow-style wallet ownership, including stonewall decoys as returned change
- Resolve Sankey input labels from prior tx / outpoint maps; classify change, self-send, receive, and Sparrow-style fake mix on charts
- Improve transaction Sankey UX: pending `?` spend status filled from network outspends, fade spent arms, mute spent labels, tap I/O to linked txs, clear spinner after labels paint
- Polish transaction details: muted `bytes` / `sats` units, Fee label, comma-formatted block height, from/to I/O counts in the header
- Improve label UX: whole label area tappable, multiline length-scaled textarea, keyboard-friendly edit flow (Return → tags, cancel/save)
- Show Sparrow-style server banners on connection tests (Electrum banner / Core subversion + sync status)
- Include transaction labels in encrypted backup/restore (custom backends and Nostr sync relays/nsec were already covered)
- Fix Nostr shared descriptor keys for Electrum seeds (`InvalidMnemonic`)
- Generate personal device Nostr keys from a BIP39 seed (NIP-06), show seed next to QR, persist `deviceMnemonic` in SQLite/backup with Fast Refresh–safe migrations

Follow-up to #447 with the remaining ownership, chart, label, backup, and Nostr device-key work.

## Test plan
- [ ] Sync an account (Electrum and/or Core RPC) and confirm owned change / decoy outputs are marked correctly in history balances
- [ ] Broadcast a stonewall (or change+decoy) send and verify optimistic confirmation keeps returned outputs in balance/UTXOs
- [ ] On draft IO preview and sign/preview charts, input nodes show prior-tx labels (not fabricated vin labels)
- [ ] On transaction detail Sankey: receive outputs are green (not outgoing); self-send only on wallet sends; spent arms fade; pending outputs show `?` then resolve via network
- [ ] Tap spent outputs / inputs to open linked txs; flow spinner clears only after chart labels are visible
- [ ] Transaction details table: raw size shows muted `bytes`, fee shows muted `sats`, block height uses commas (e.g. `850,123`), header shows from N inputs / to M outputs
- [ ] Tap anywhere on the label block to edit; label field is multiline and shrinks text as the label grows; Return moves to tags; Cancel discards
- [ ] Network settings connection test shows Electrum banner or Core subversion/status when available
- [ ] Encrypted backup/restore keeps transaction labels, custom backends, and Nostr sync relays/nsec
- [ ] Electrum-seed wallet: Nostr Shared descriptor keys generate without `InvalidMnemonic`
- [ ] Personal device keys: New NSEC creates a 12-word seed; Show seed appears next to Show QR; seed restores from backup after regenerate + save
- [ ] Unit tests: `walletOwnership`, `sankeyInputLabel`, `sankeyOutputLabel`, `stonewall`, `transactionOutspends`, `recoverBackup`, `bip39` Electrum descriptors, `nostrIdentity` device seed


apps/mobile/utils/walletOwnership.ts
```diff
@@ -0,0 +1,198 @@
...
+  const { internalAddresses, ownAddresses } =
+    getAccountAddressSets(accountAddresses)
+
+  if (ownAddresses.size === 0) {
+    return transactions
+  }
+
+  return transactions.map((transaction) => {
+    const vout = transaction.vout.map((output) => {
+      const address = output.address.trim()
```
Psycarlo: Unguarded `output.address.trim()` will crash on addressless outputs.

Fix: `output.address?.trim()` at all new sites (the `if (!address)` check already handles the falsy result), matching the sibling function.

Psycarlo (changes requested):

Psycarlo:
## 🔴 HIGH — unguarded `output.address.trim()` crashes on addressless outputs

### Root cause

`transaction.vout[].address` is typed `string` but is `undefined` at runtime for addressless outputs (OP_RETURN / non-standard scripts):

- `apps/mobile/api/electrum.ts:388-392` — pushes `generateOutputScriptAddress()` raw (returns `undefined` for OP_RETURN).
- `apps/mobile/api/bdk.ts:876` — defensively writes `output.address || ''`.
- `walletOwnership.ts` sibling `collectTransactionOutputAddresses` already uses `output.address?.trim()`.

So the codebase already treats `address` as possibly-falsy. New code drops the guard → `TypeError` → screen crash.

### Fix

Use `output.address?.trim()` everywhere below. The existing `if (!address)` checks already handle the falsy result.

---

### `apps/mobile/utils/walletOwnership.ts` (new file) — highest blast radius

Runs in the account-overview render path (`TotalTransactions`, `SatsInMempool`, `TxDetails`). One tx with an OP_RETURN output crashes the whole account screen.

**Line 127** — vout `.map`, change-tagging:
> HIGH: `output.address` is `undefined` at runtime for OP_RETURN / non-standard outputs (electrum.ts pushes it raw; bdk.ts guards with `|| ''`; the sibling `collectTransactionOutputAddresses` already uses `?.trim()`). `.trim()` on undefined throws `TypeError` and crashes the account-overview render path (`TotalTransactions` / `SatsInMempool` / `TxDetails`). Use `output.address?.trim()` — the `if (!address ...)` check below already handles the falsy result.

**Line 144** — returned/external sum loop:
> HIGH: same issue — `output.address` can be `undefined`, so `.trim()` throws. Use `output.address?.trim()`; the `if (address && ownAddresses.has(address))` guard already handles falsy.

---

### `apps/mobile/components/SSTransactionChart.tsx`

Crashes the transaction-detail Sankey.

**Line 265** — `pendingOutspendOutputs` loop (**new**):
> HIGH: `output.address` is optional at runtime — electrum pushes `undefined` for OP_RETURN / non-standard outputs (see electrum.ts `generateOutputScriptAddress`), and bdk.ts writes `|| ''`. `.trim()` on undefined throws `TypeError` and crashes the tx-detail chart. Use `output.address?.trim()` — the `if (!address) continue` below already handles the falsy result.

**Line 395** — `classifyChartOutputs` `to:` (**new**):
> HIGH: same issue — `output.address` can be `undefined` at runtime (OP_RETURN), so `.trim()` throws. Use `output.address?.trim() ?? ''` here since it feeds the `to` field.

**Line 405** — `const outputAddress` (**pre-existing**, not introduced by this PR):
> Same latent crash as above (pre-existing). Since you're touching this function, worth guarding too: `output.address?.trim() ?? ''`. An addressless output crashes the whole Sankey render.

---

## Below-bar notes (not blockers)

- **MEDIUM** — inline settings banner (`lastBanner` / `lastServerBanner` in `settings/network/[network].tsx` & `server.tsx`) renders the full server-supplied banner uncapped, unlike the toast path (`formatBannerForToast`). A hostile/compromised Electrum server could return a huge banner and bloat the settings view. Fix: reuse `formatBannerForToast` (or a line/char cap) for inline display too.
- **INFO** — `nostr_device_mnemonic` stored as plaintext `TEXT` (schema V6). It's the Nostr device identity seed, not the Bitcoin wallet seed, and mirrors the existing plaintext `deviceNsec` in the same table — no new exposure class. If hardening Nostr secret storage later, move both fields together.

apps/mobile/utils/walletOwnership.ts
pedromvpg: Fixed in df74ad5 — switched to `output.address?.trim()` in `walletOwnership.ts` and the Sankey chart sites (including the pre-existing one). Also capped the inline server banners with `formatBannerForToast`.

pedromvpg:
Addressed in df74ad5:

1. **HIGH** — `output.address?.trim()` in `walletOwnership.ts` (change-tagging + returned/external sum) and `SSTransactionChart.tsx` (pending outspends, `classifyChartOutputs`, and the pre-existing `outputAddress` path)
2. **MEDIUM** — inline settings banners now go through `formatBannerForToast`
3. **INFO** — noted on plaintext `nostr_device_mnemonic`; left as-is since it matches existing `deviceNsec` storage

Added a regression test for addressless OP_RETURN outputs in `walletOwnership.test.ts`.

apps/mobile/utils/walletOwnership.ts
pedromvpg: Fixed in df74ad5 — switched to `output.address?.trim()` in `walletOwnership.ts` and the Sankey chart sites (including the pre-existing one). Also capped the inline server banners with `formatBannerForToast`.

pedromvpg:
Addressed in df74ad5:

1. **HIGH** — `output.address?.trim()` in `walletOwnership.ts` (change-tagging + returned/external sum) and `SSTransactionChart.tsx` (pending outspends, `classifyChartOutputs`, and the pre-existing `outputAddress` path)
2. **MEDIUM** — inline settings banners now go through `formatBannerForToast`
3. **INFO** — noted on plaintext `nostr_device_mnemonic`; left as-is since it matches existing `deviceNsec` storage

Added a regression test for addressless OP_RETURN outputs in `walletOwnership.test.ts`.

Psycarlo (approved):

PR #452, opened 2026-07-19
feat: toggle visibility of UI-heavy charts

v4v2:
Add option in settings > features > chart to toggle visibility of Transaction flow chart & utxo bubbles.

These charts are heavy and may slow down the app in some occasions or in some phones.

By default, the charts are shown.


Psycarlo:
@pedromvpg What do you think?

Psycarlo (approved):

PR #453, opened 2026-07-19
ui: add an actual bitcoin icon

v4v2:
Previously it was just plain circle icon instead of bitcoin icon LOL.

The icon is displayed in the signer menu (bitcoin, lightniung, ark, liquid, ecash, nostr).

The icon was taken from Lucidev, as per our standards.


Psycarlo:
Let's keep things more abstract. Keep like we have now.

PR #455, opened 2026-07-22
address private key & message signature - BIP 137 & BIP 322

v4v2:
This adds the features:

- [x] view bitcoin address key material
- [x] ensure code related to address key material can be used elsewhere
- [x] add UI to sign arbitrary messages using a bitcoin address' private key 
- [x] add UI to verify integrity of messages signed from bitcoin address
- [x] implement BIP 137 for single signature (sign & verify) 
- [x] implement BIP 322 for P2TR scripts (sign & verify)
- [x] check BIP137 implementation works (compare with Sparrow)
- [x] check BIP322 (P2TPR) implementation works (compare with Sparrow)
- [x] add tests


apps/mobile/tests/unit/utils/bip137Samples.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * BIP-137 sign/verify test vectors.
```
Psycarlo, line 2: Why using `_` for file names now?

apps/mobile/utils/bitcoin.ts
```diff
@@ -264,6 +330,47 @@ export function getDerivationPathFromScriptVersion(
   }
 }
 
+export function getAccountDerivationPath(
```
Psycarlo, line 333: `getAccountDerivationPath` ignores `key.derivationPath`

apps/mobile/utils/bitcoin.ts
```diff
@@ -264,6 +330,47 @@ export function getDerivationPathFromScriptVersion(
...
+
+/**
+ * Full derivation path for a specific address. Addresses built from BDK's
+ * peekAddress() (see api/bdk.ts#getWalletAddresses) don't carry their own
+ * derivationPath, so this reconstructs it from the account path plus the
+ * change level (0=external, 1=internal) and index. IMPORTANT: the change
+ * level must be included - `${accountPath}/${index}` alone derives a
+ * completely different (wrong) key, since it's missing a whole path level.
+ */
+export function getAddressDerivationPath(
```
Psycarlo, line 360: Inconsistent `m/` prefix on the returned path.

We should always return with the prefix, or without it, for consistency

apps/mobile/utils/key.ts
```diff
@@ -80,3 +88,42 @@ export function extractPublicKeyFromKey(
 
   return ''
 }
+
+export function getAddressKeyPair(
```
Psycarlo, line 92: **Electrum seeds should be rejected explicitly, not left to a checksum coincidence**

The mnemonic branch calls `mnemonicToSeed` (BIP-39 PBKDF2, salt `"mnemonic"+passphrase`), but Electrum accounts were built with `mnemonicToSeedElectrum` (salt `"electrum"+passphrase`, `utils/bip39.ts:98`) — a different seed, so a different key tree.

In practice this is *mostly* safe by accident: Electrum seeds are generated for Electrum's HMAC version prefix, not the BIP-39 checksum, so `Mnemonic.fromString` throws and the `try/catch` returns `null` → UI shows "unavailable". I measured ~38/40 Electrum seeds failing the BIP-39 checksum, matching the expected 15/16.

The remaining ~1-in-16 pass both checks. There `fromString` succeeds and we derive a wrong key silently. Cheap to close explicitly:

```ts
if (secret.mnemonic) {
  if (detectElectrumSeed(secret.mnemonic)) {
    return null // Electrum seeds need mnemonicToSeedElectrum (async)
  }
  const seed = mnemonicToSeed(secret.mnemonic, secret.passphrase)
  return getAddressKeyPairFromSeed(seed, address.derivationPath)
}
```

Returning `null` is fine for now given `mnemonicToSeedElectrum` is async and this function is sync — `keyUnavailable` already renders `address.details.key.unavailable`. Low severity on its own; noting it since "silently wrong private key" is a bad failure mode even at 1/16.

Psycarlo (changes requested):

Psycarlo (approved):

PR #464, opened 2026-07-30
refactor PIN and encryption/decryption utils

### LINKED ISSUE #463: refactor set PIN
There are dozens of places in our codebase calling `setItem(PIN_KEY, hashedPin)` and `getItem`, `DEFAULT_PIN`, `PIN_KEY`.
TSX Pages shall not directly make calls to `getItem`, which is more like a low-level access to secure store.
Instead, everyone should call an intermediator `setPin(string) => void` which will handle hashing the PIN and storing it in secure store.

v4v2:
This is meant to avoid directly calling `getItem(PIN_KEY)`, `setItem(hashedPin)`, `aesDecrypt(encryptedString, pin, iv)`, `aesEncrypt(plainString, pin, iv)` because they are directly accessing secure store instead of using utils designed for this.

- [x] replace getItem(PIN_KEY) -> getPin() whenever possible
- [x] replace generateSalt() + encryptPin(pin + salt) + setItem(PIN_KEY, pin) -> setPin(pin, PIN_KEY) whenever possible
- [x] replace directly calls to aesEncrypt/aesDecrypt whenever possible

Manual tests:
- [x] enter pin
- [x] change pin
- [x] decryption of key secret
- [x] reencryption of key secrets when pin changes
- [x] set up duress pin
- [x] duress pin working

Fixes #463.


Psycarlo:
## Review

Consolidating `getPin`/`setPin` into `utils/pin.ts` is the right direction, and replacing the hand-rolled `aesDecrypt` + `JSON.parse` blocks with `decryptAccountKeySecret`/`decryptKeySecret` removes real duplication. Three blockers below, all in the duress-PIN path or the refactored effects.

---

### 🔴 Duress PIN breaks for existing users — no migration

`SALT_KEY_DURESS` is new in this PR. On `master`, the duress PIN was hashed with the shared `SALT_KEY` (both `store/auth.ts:setDuressPin` and `app/setDuressPin.tsx` wrote it that way).

`components/SSPinAuth.tsx`:
```ts
const saltDuress = await getItem(SALT_KEY_DURESS)
const hashedInputDuress = saltDuress ? await pbkdf2Encrypt(inputPin, saltDuress) : ''

if (duressPinEnabled && hashedInputDuress === hashedDuressPin) {
```

Existing users have `DURESS_PIN_KEY` set but no `SALT_KEY_DURESS`. So `hashedInputDuress` is `''`, never matches, and the duress PIN silently stops working — it falls through and is treated as a wrong PIN.

The feature fails silently in exactly the scenario it exists for. Needs a fallback to `SALT_KEY` when `saltDuress` is null, plus a re-derive on next successful duress auth.

---

### 🔴 `setDuressPin` validates a hash it never stores

`app/setDuressPin.tsx`:
```ts
const salt = await getItem(SALT_KEY)
const regularPinHashed = await getPin()
...
const encryptedDuressPin = await pbkdf2Encrypt(pin, salt)   // hashed with SALT_KEY
if (regularPinHashed === encryptedDuressPin) { ...reject... }
await setPin(pin, DURESS_PIN_KEY)                            // generates its own new salt
```

The comparison hashes with `SALT_KEY`, but `setPin` internally calls `generateSalt()` and writes to `SALT_KEY_DURESS`. So the value that gets compared is never the value that gets stored.

The comparison itself still works on `master`'s terms (same salt on both sides), so the guard isn't strictly dead — but the stored duress hash is now unrelated to the one that was checked. Simpler and correct: compare the raw PIN strings.

---

### 🔴 `try/catch/finally` wrapping un-awaited async calls

Three screens moved the async body out of the effect but left the try/catch behind:

- `.../account/[id]/settings/export/descriptor.tsx`
- `.../account/[id]/settings/export/publicKey.tsx`
- `.../account/add/multiSig/export/publicKey.tsx`

```ts
useEffect(() => {
  setIsLoading(true)
  try {
    getPublicKey()          // async, not awaited
  } catch {
    toast.error(...)        // never fires
  } finally {
    setIsLoading(false)     // runs immediately
  }
}, [...])
```

Two bugs. `catch` cannot catch a rejected promise, so decryption failures become unhandled rejections with no toast — a regression, since the old code caught them. And `finally` clears the loading flag synchronously, before decryption completes, so the spinner ends early and the screen renders empty.

`descriptor.tsx` is worst: `getDescriptor()` now awaits `decryptAccountKeySecret`, which throws when the PIN is unavailable, and nothing catches it.

Fix: move the try/catch inside the async function, or chain `.catch().finally()`.

---

### 🟡 `getPin(DEFAULT_PIN_KEY)` returns a plaintext PIN

The `crypto.ts` docstring deleted in this PR said *"Never returns a plaintext default PIN — callers that need encryption must use the stored digest."* The new `utils/pin.ts` does exactly that:

```ts
async function getPin(pinType = PIN_KEY): Promise<string> {
  if (pinType === DEFAULT_PIN_KEY) {
    return DEFAULT_PIN
  }
```

No caller passes `DEFAULT_PIN_KEY` today, so it's latent. But it returns plaintext `'2121'` where every caller expects a PBKDF2 digest, and it isn't gated on `__DEV__`.

The `PinType` union is also wrong:
```ts
type PinType = typeof PIN_KEY | typeof DURESS_PIN_KEY | typeof DEFAULT_PIN
```
The last member is `DEFAULT_PIN` (the PIN value), not `DEFAULT_PIN_KEY` (the storage key). All three are `string`, so the union widens to `string` and gives no type safety — `setPin(pin, 'anything')` compiles.

Drop the branch, or fix the name and gate on `__DEV__`.

---

### 🟡 `secureWipe` deletes are no longer error-tolerant

`utils/secureWipe.ts` dropped the `.catch(() => undefined)`:
```ts
await deleteItem(DURESS_PIN_KEY)
await deleteItem(SALT_KEY_DURESS)
```

Every other delete in this function is still `.catch`-guarded. `deleteItemAsync` rejects when deletion fails, so if the first rejects the second never runs and `SALT_KEY_DURESS` survives the wipe. The caller in `SSPinAuth` swallows it and proceeds to unlock, matching the "best-effort" comment — but the second delete is skipped either way. Keep the `.catch` on both for consistency with the rest of the function.

---

### 🟡 `performRecoverOverwrite` loses definite-assignment checking

`utils/recoverBackup.ts`: `let pin: string` → `let pin = ''`.

Behavior is unchanged today, since the `catch` returns early. But the initializer removes TypeScript's protection: if that `catch` is ever edited to not return, `pin` silently becomes `''` and every secret is encrypted under an empty key, with no compiler complaint. Revert to `let pin: string`.

---

### 🟢 Minor

- multiSig `publicKey.tsx`: the dep array still lists `convertPublicKeyFormat`, now a plain function recreated every render. Inert behind the disable comment, but misleading — drop it.
- `store/auth.ts:setDuressPin` now delegates to `setPin(pin, DURESS_PIN_KEY)`, but nothing calls it — `settings/security.tsx` routes to `/setDuressPin` instead. Dead code worth removing while you're here.
- `BackupJsonSchema` uses `z.codec`, which requires zod ≥4. `apps/mobile/node_modules/zod` resolves to 4.4.3, so this works. Repo root has 3.25.76 — worth knowing if anything ever hoists.

---

### Verdict

Direction is good, but not mergeable as-is. The duress-PIN migration break is the serious one: existing duress PINs stop working with no warning and no migration path. The un-awaited try/catch trio is a straightforward fix.

All findings are from reading the diff against `master` — I didn't run the test suite or the app. The duress migration break in particular is worth confirming on a device with a pre-existing duress PIN.

v4v2 (commented):
@psycarlos review again. Dismiss item about duress-PIN migration since in this particular scenario is ok.

Psycarlo (approved):

Psycarlo:
## Blockers

All three blockers from the previous round are fixed — `await` added inside the try/catch in the export screens, `setDuressPin` now uses `checkPinEqual`, `PinType` corrected. One new blocker from this round's abstractions.

---

### 🔴 `useVerifyConnection` can leak NetInfo listeners

On `master`, the listener was registered **synchronously** and `unsubscribe` captured in the closure, so cleanup always had something to call:

```ts
;(async () => {
  const state = await NetInfo.fetch()
  if (cancelled) return
  isConnectionAvailable.current = ...
  refetch()
})()

const unsubscribe = NetInfo.addEventListener(...)   // synchronous

return () => {
  cancelled = true
  unsubscribe()
}
```

Now the listener is created **after** the await and returned as the cleanup value:

```ts
useAsyncEffect(async () => {
  const state = await NetInfo.fetch()
  isConnectionAvailable.current = ...
  refetch()

  return NetInfo.addEventListener(...)   // only exists after fetch resolves
}, [config.connectionMode, refetch])
```

If the effect re-runs before that fetch resolves, `useAsyncEffect` runs cleanup while `cleanup` is still `undefined`. Run 1's listener resolves afterwards and is written into a variable nobody reads again — never unsubscribed. Each leaked listener keeps calling `refetch()`.

Deps are `config.connectionMode` and `refetch`, so toggling connection mode while a fetch is in flight triggers it.

To be fair to the hook: the pure-unmount case *is* handled — `result?.()` tears down a listener that arrives after the effect goes inactive. The leak is specifically the **re-run** case.

Two contributing defects in `useAsyncEffect` itself:

```ts
callback().then((result) => {
  if (active) {
    cleanup = result
  } else {
    result?.()
  }
})
```

1. **No `.catch`.** A rejecting callback becomes an unhandled promise rejection. There's no global rejection handler in the app, so it's a redbox in dev and silently dropped in production. The export screens are safe (they try/catch internally); `useVerifyConnection` is not — `await NetInfo.fetch()` rejects straight into the hook.

2. **No mid-flight cancellation.** The `active` flag gates only whether cleanup runs; it does not stop the callback body executing after unmount. `useVerifyConnection` previously had an explicit `cancelled` check after its await and lost it, so a late-resolving fetch still writes `isConnectionAvailable.current` and calls `refetch()` on an unmounted component.

**Fix either side:** give the hook a `.catch` and pass an `isActive()` predicate to the callback, or leave `useVerifyConnection` on a manual `useEffect`.

PR #467, opened 2026-07-31
feat: Payjoin, UTXO tooling, explorer, and UI polish

pedromvpg:
## Summary
- **Payjoin:** add `react-native-payjoin` (UniFFI) and wire BIP77/BIP78 send & receive — directory vs manual coordination modes, session TTL/resume, mixed-ownership UI, send preview/negotiation, offline PSBT handoff, Signet live diagnostics, and Maestro e2e flows.
- **UTXO selection & coin control:** sort, group, filter (script/tag), exclude, and nested bubble layouts; wait for fee estimate before URI auto-select; refine filter overlay and multisig key actions.
- **Settings & transport:** directory vs manual Payjoin mode settings; reusable PSBT import/export transport UI; improved custom network URL paste parsing; shared backup filename helper; persist signed tx drafts across restarts.
- **Seed / entropy UX:** drawing entropy input and polish for coin/dice seed screens.
- **Explorer & fees:** node-first explorer (address, block, tx, chain, halving, mempool, network), RPC/Esplora helpers, block hex view/dump/decode, 30-day price chart, size-weighted mempool histogram, fractional sat/vB on sankey, special-output tagging, confirmation backlog estimates.
- **UI polish:** thinner toolbar icon strokes, slightly smaller toggles, compact empty label/tag placeholders, account layout/sync timestamp hardening, lint/format cleanup.

## Test plan
- [ ] Build/run Android with Payjoin native module linked
- [ ] Payjoin receive (BIP77 invoice, QR/URI, resume after restart, expiry/TTL)
- [ ] Payjoin send (directory URI → preview → negotiate → sign → broadcast)
- [ ] Manual/offline mode: PSBT import/export contribute/finalize between wallets
- [ ] Settings: toggle directory vs manual Payjoin coordination; paste custom network URLs
- [ ] UTXO list: sort/group/filter by script & tags, exclude coins, nested bubbles
- [ ] Seed flow: drawing entropy + coin/dice screens
- [ ] Signed tx draft survives app restart; backup filename uses shared helper
- [ ] Explorer: address/block/tx search & details, mempool backlog, price chart, block hex views
- [ ] Maestro Payjoin e2e flows on Android
- [ ] Optional: Signet live roundtrip developer diagnostics


Psycarlo:
## 1. Online receiver skips both BIP78 safety checks — blocker

`apps/mobile/api/payjoinNative.ts:692-695`

```ts
.checkInputsNotOwned({ callback: () => false })
.checkNoInputsSeenBefore({ callback: () => false })
```

Both stubbed. `() => false` means "not mine" and "never seen" unconditionally. This is the live BIP77 directory path (`contributeReceiver` ← `receiverContributeAndFinalize` ← `api/payjoin.ts:1478`).

- `checkNoInputsSeenBefore` disabled → probing attack: sender repeatedly opens and aborts payjoins to enumerate receiver UTXOs. JS-side `hasSeenInput` (`api/payjoin.ts:1446`) softens but deliberately falls through to reusing a seen input, so it is not equivalent.
- `checkInputsNotOwned` disabled → malicious sender can submit an original PSBT already spending the receiver's own coins, and the receiver signs it.

The correct data exists and is already plumbed on the manual path: `receiverManualContribute` (`payjoinNative.ts:734`) takes `ownedScriptsHex`, `seenOutpoints`, `ownedOutpoints`. The online function signature has no parameter for them; at `payjoin.ts:1478` `params.callbacks` is in scope with `isScriptOwned` and `hasSeenInput` already built, and is not passed.

**Fix:** give `receiverContributeAndFinalize` the same three arguments and wire the real callbacks through.

## 2. TS validator accepts substituted payee when `pjos=1`

`apps/mobile/utils/payjoinValidate.ts:96-113`

Probe results against real code:

| Probe | Result |
|---|---|
| pjos=1, payee script swapped for attacker's, amount kept | `{"ok":true}` — accepted |
| pjos=1, sender change 49 000 → 1 sat | `{"ok":true}` — accepted |
| pjos=0, outputs reordered | rejected (correct) |
| pjos=1, payment reduced to 10 000 | rejected (correct) |

The amount check sums *every* non-owned output instead of the output paying the intended payee script, so an attacker's substituted output still counts toward `paidToReceiver`. Separately, nothing bounds how far sender change may decrease, so it can be drained to fee.

**Scope, stated precisely:** on BIP77/v2 the PSBT from `senderProcessResponse` is already validated by rust-payjoin, so this is defense-in-depth there. It is the *sole* validator on the BIP78 pure-HTTP fallback (`api/payjoin.ts:482-507`), reached when native is unavailable or the native branch throws (line 477 swallows and falls through).

**Fix:** carry the intended payee script through and assert that specific output pays ≥ `paymentAmountSats`; bound sender-change decrease. Note `tests/unit/utils/payjoinValidate.test.ts:178` encodes current behavior and will not catch this.

## 3. `usePayjoinSender` is dead code

`apps/mobile/hooks/usePayjoinSender.ts` — zero callers repo-wide. `signTransaction.tsx` calls `sendPayjoin` / `startBip77Send` / `pollBip77Send` directly with its own `buildCallbacks()`. Two divergent copies of callback construction: a fix to one silently misses the other. Delete the hook or migrate the screen onto it.

pedromvpg:
Thanks @Psycarlo — all three addressed in `feature/payjoin` (58ec167, 2579fa8, 5efc626):

**1. Online receiver BIP78 checks** — replaced both `() => false` stubs in `contributeReceiver`. Rather than re-deriving arrays, I thread the receiver's own `isScriptOwned` / `hasSeenInput` predicates (sync-coerced) from `params.callbacks` through `receiverContributeAndFinalize` into the real `checkInputsNotOwned` / `checkNoInputsSeenBefore` callbacks — the data you flagged as already-built-but-not-passed. Checks are consumed only on the contribute pass. Added a unit test that spies on the native adapter to assert the predicates are wired and answer correctly.

**2. Validator payee anchor + change bound** — carried the sender's intended payee scripts (`paymentScriptsHex`) into `validatePayjoinProposal` and replaced the "sum every non-owned output" logic with:
- each payee output must survive at its original index with value ≥ original (and ≥ requested amount for a single payee), **even when `pjos=1`** — so a swapped payee is rejected;
- sender-change decrease is bounded by the receiver's net contribution (new inputs − new outputs), rejecting drains to fee.

Plumbed `outputScriptsHex` through `sendBip78`/`sendBip77`, the BIP77 start/poll path, `finalizeSenderProposal`, and `applyManualSenderProposal`. Replaced the misleading `payjoinValidate.test.ts:178` case and added adversarial coverage (swapped payee, drained change, reduced amount) plus a legitimate-accept case. Your probe table now returns `{"ok":false}` for the two accepted-attack rows.

**3. `usePayjoinSender`** — deleted (zero callers).

Verification: full unit suite green (1602), 5 payjoin suites green (36), lint/format clean.

One caveat: the ubrn-generated TS bindings aren't in the checkout, so the exact callback arg shapes in `contributeReceiver` (`script: ArrayBuffer`, `outpoint: {txid, vout}`) are based on the in-repo `identifyReceiverOutputs` shape + payjoin-ffi's `OutPoint` and need a native rebuild + on-device signet round-trip to fully validate.

Psycarlo (approved):

Psycarlo (approved):

PR #469, opened 2026-08-03
fix: security hardening for PIN, paths, seeds, WYSIWYS, and duress wipe

pedromvpg:
## Summary
- Gate `skipPin` / `DEFAULT_PIN` to development builds so production never uses plaintext `2121` as encryption key material or lock-screen bypass.
- Sanitize export filenames and ark datadir IDs to block path traversal.
- Require PIN before revealing ecash seed words (aligned with Bitcoin seed reveal).
- Derive sign-confirmation I/O from the PSBT itself (WYSIWYS) instead of mutable builder state.
- Harden duress PIN wipe to clear key secrets, ecash/ark, Nostr, LND config, and RPC credentials.

## Test plan
- [ ] Fresh install: set PIN flow has no “set PIN later” in a production/release build; `__DEV__` still allows skip for local demos
- [ ] With skipPin enabled in a debug build, encrypt/decrypt accounts still uses SecureStore PIN material (not plaintext `2121`)
- [ ] Ecash account → Settings → Seed: reveal prompts for PIN; wrong PIN fails; success shows words
- [ ] Export descriptor/labels with an account name containing `/`, `..`, or spaces — shared filename is sanitized
- [ ] Import/load a PSBT where builder store I/O is empty or stale — sign screen chart matches PSBT outputs/amounts
- [ ] Set duress PIN, enter it at lock — accounts, ecash/ark, Nostr identities, LND config, and RPC creds are gone; duress PIN evidence removed
- [ ] `pnpm run check` and `pnpm --filter @satsigner/mobile typecheck` pass
- [ ] `pnpm --filter @satsigner/mobile test:unit` passes


Psycarlo:
## 1. Duress wipe leaves RPC credentials in `customServers` (contradicts the PR's own test plan)

`secureWipeAllWalletData` strips `rpcCredentials` only from the **active** server config per network (`apps/mobile/utils/secureWipe.ts`, the `BLOCKCHAIN_NETWORKS` loop):

```ts
const { configs, updateServer } = useBlockchainStore.getState()
for (const network of BLOCKCHAIN_NETWORKS) {
  const { server } = configs[network]
  if (!server.rpcCredentials) continue
  const { rpcCredentials: _removed, ...rest } = server
  updateServer(network, rest)
}
```

But saving an RPC server also stores it — credentials included — in the `customServers` array via `addCustomServer(server)` (`settings/network/[network].tsx:308`, built with `rpcCredentials` at line 228). That array is persisted by the blockchain store's `partialize` (`store/blockchain.ts:159`) into **unencrypted** MMKV (`storage/mmkv.ts:6` — `createMMKV({ id: 'mmkv.satsigner' })`, no `encryptionKey`).

After a duress wipe, plaintext node username/password survive on disk. The schema's own comment (`types/settings/blockchain.ts`) notes these credentials can grant full RPC control including spend on a hot node wallet. The PR test plan says "RPC creds are gone" — not true for the custom server list.

**Fix:** also strip `rpcCredentials` from every entry in `customServers` (or drop custom RPC servers entirely) inside `secureWipeAllWalletData`.

---

## 2. Duress wipe leaves transaction-builder drafts (wallet data) in MMKV

The `satsigner-transaction-builder` persist key stores per-account drafts containing full UTXO objects (`txid`, `vout`, `value`, `addressTo`, `label`) and output destinations/amounts (`store/transactionBuilder.ts:370-373`):

```ts
partialize: (state) => ({
  accountId: state.accountId,
  drafts: state.drafts
})
```

`secureWipeAllWalletData` never touches `useTransactionBuilderStore`. After a duress wipe, recoverable evidence of the wiped wallets (UTXO set, destination addresses, amounts, labels) remains in unencrypted MMKV — directly against the stated goal of "acting as if the duress pin was the true pin."

Same class, lower value: the `satsigner-nostr` store (profiles, group members keyed by account id) and the `nostr:follows:*` cache keys (`storage/mmkv.ts:24-26`) also survive.

**Fix:** clear the transaction-builder store (and ideally the nostr sync store + follow caches) in `secureWipeAllWalletData`.

---

## 3. One unguarded rejection aborts the entire wipe silently

The first deletion batch in `secureWipeAllWalletData` has no error handling, unlike the ecash/ark batches right below it which use `.catch(() => undefined)`:

```ts
await Promise.all(
  accounts.map((account) =>
    deleteAllKeySecrets(account.id, account.keys.length)  // no .catch
  )
)
```

If any SecureStore delete rejects, the function throws **before any store is cleared and before `DURESS_PIN_KEY` is deleted**. The rejection then propagates unhandled through `handleFillEnded` in `SSPinAuth.tsx` (the `await secureWipeAllWalletData()` call has no try/catch), so the user stays at the lock screen with nothing wiped and the duress-PIN evidence intact.

**Fix:** a duress wipe must be best-effort all the way down — wrap this batch like the others, and consider try/catch around the whole call in `SSPinAuth` so navigation always proceeds.

---

## 4. Migration gap: existing "Set PIN later" users can no longer change their PIN

On master, production builds showed "Set PIN later," which set `skipPin = true` and silently set the real PIN to `DEFAULT_PIN` (`2121`). This PR force-clears the flag on rehydrate (`store/auth.ts`):

```ts
onRehydrateStorage: () => (state) => {
  if (!__DEV__ && state?.skipPin) {
    state.skipPin = false
  }
}
```

Consequence for that existing cohort: Settings → Security → PIN now starts at the **verify** stage (`setPin.tsx:55` — `fromSettings && !skipPin ? 'verify' : 'set'`), demanding a current PIN the user never knowingly chose. They cannot set a real PIN unless they guess `2121`. Funds are not locked (`requiresAuth` remains false), but PIN setup is permanently bricked for them, and their key material stays derived from `2121`.

**Fix:** one-time migration — if `skipPin` was true on rehydrate, either route these users into the `set` stage directly, or detect `validatePin(DEFAULT_PIN)` success and treat the PIN as unset.

---

## 5. Pre-existing, but this PR rewrites the exact block: duress success loops back to the lock screen

The duress branch in `SSPinAuth.tsx` never clears `lockTriggered` or sets `justUnlocked`:

```ts
if (duressPinEnabled && hashedInput === hashedDuressPin) {
  await secureWipeAllWalletData()
  router.dismissAll()
  router.push('/')
  return
}
```

After `router.push('/')`, the guard in `app/(authenticated)/_layout.tsx:203` (`requiresAuth && lockTriggered && !skipPin`) redirects straight back to `/unlock` — where the duress PIN no longer works (its key was just deleted). The user is stranded and the "duress PIN looks like the real PIN" deception is broken. Behavior is unchanged from master, but since this commit rewrites the block, the fix is cheap here: `setLockTriggered(false)`, `setJustUnlocked(true)`, `resetPinTries()` before navigating.

pedromvpg:
Superseded by #472 — same commits, head branch now hosted on the upstream satsigner/satsigner repo instead of my fork.

PR #470, opened 2026-08-03
fix: encrypt secrets at rest and harden skipPin / LNURL

pedromvpg:
## Summary
- PIN-encrypt Nostr secrets (identity + per-account) in SecureStore; stop persisting nsec/mnemonic in MMKV/SQLite
- PIN-encrypt LND macaroon/cert and Bitcoin Core RPC credentials the same way; strip them from MMKV persistence; migrate/hydrate on unlock; re-encrypt on PIN change; wipe on duress
- Remove hardcoded `DEFAULT_PIN = '2121'`. `skipPin` only skips the lock screen in `__DEV__`; encryption uses a random ephemeral passphrase (PBKDF2 digest) via `enableDevSkipPin()`
- Hide developer settings outside `__DEV__` (settings entry + deep-link redirect)
- Require HTTPS for LNURL fetch/callback URLs; set Android `usesCleartextTraffic` only in development

## skipPin design
Hardcoding a default PIN is wrong — it would be a known decryption key. The better model:
1. `skipPin` = UI lock bypass (dev-only, forced off in production rehydrate)
2. Encryption key material = random 32-byte ephemeral secret → PBKDF2 → SecureStore (same path as a real PIN)
3. The ephemeral plaintext is never persisted; only the digest is kept (matching normal PIN storage)

## Test plan
- [ ] Fresh install: set a real PIN; unlock; Nostr / LND / RPC secrets round-trip after kill+relaunch
- [ ] Change PIN: secrets still decrypt after re-encrypt
- [ ] Duress PIN: LND/RPC/Nostr secrets wiped with other wallet data
- [ ] `__DEV__`: "Skip PIN" / set PIN later creates encryption material without using `2121`; lock screen is skipped
- [ ] Release/production build: developer settings card absent; `skipPin` ignored if somehow persisted
- [ ] LNURL: `http://` endpoints rejected; HTTPS still works
- [ ] Existing installs with plaintext LND/RPC in MMKV migrate into SecureStore on next unlock


pedromvpg:
Superseded by #471 — same commits, head branch now hosted on the upstream satsigner/satsigner repo instead of my fork.

PR #472, opened 2026-08-04
fix: security hardening for PIN, paths, seeds, WYSIWYS, and duress wipe

pedromvpg:
## Summary
- Gate `skipPin` / `DEFAULT_PIN` to development builds so production never uses plaintext `2121` as encryption key material or lock-screen bypass.
- Sanitize export filenames and ark datadir IDs to block path traversal.
- Require PIN before revealing ecash seed words (aligned with Bitcoin seed reveal).
- Derive sign-confirmation I/O from the PSBT itself (WYSIWYS) instead of mutable builder state.
- Harden duress PIN wipe to clear key secrets, ecash/ark, Nostr, LND config, and RPC credentials.

## Test plan
- [ ] Fresh install: set PIN flow has no “set PIN later” in a production/release build; `__DEV__` still allows skip for local demos
- [ ] With skipPin enabled in a debug build, encrypt/decrypt accounts still uses SecureStore PIN material (not plaintext `2121`)
- [ ] Ecash account → Settings → Seed: reveal prompts for PIN; wrong PIN fails; success shows words
- [ ] Export descriptor/labels with an account name containing `/`, `..`, or spaces — shared filename is sanitized
- [ ] Import/load a PSBT where builder store I/O is empty or stale — sign screen chart matches PSBT outputs/amounts
- [ ] Set duress PIN, enter it at lock — accounts, ecash/ark, Nostr identities, LND config, and RPC creds are gone; duress PIN evidence removed
- [ ] `pnpm run check` and `pnpm --filter @satsigner/mobile typecheck` pass
- [ ] `pnpm --filter @satsigner/mobile test:unit` passes

---
_Supersedes #469 (same commits, branch now hosted on the upstream repo)._


pedromvpg:
Thanks for the thorough review — all five addressed in `12f5c55ed`.

**1. RPC creds in `customServers`** — Added `stripAllRpcCredentials()` to the blockchain store; it strips `rpcCredentials` from every network config **and** every `customServers` entry. `secureWipeAllWalletData` now calls it in place of the active-server-only loop.

**2. Tx-builder drafts + nostr data** — `secureWipeAllWalletData` now also clears the transaction-builder store (new `clearAllData()` that resets *all* drafts, not just the active account), calls `useNostrStore.clearAllNostrState()`, and clears the `nostr:follows:*` cache keys via a new `clearNostrFollowCaches()` mmkv helper.

**3. Unguarded rejection aborts the wipe** — The first `deleteAllKeySecrets` batch is now `.catch(() => undefined)` like the ecash/ark batches, and the `secureWipeAllWalletData()` call in `SSPinAuth` is wrapped in try/catch, so the wipe is best-effort all the way down and `DURESS_PIN_KEY` is always removed before navigating.

**4. "Set PIN later" migration gap** — Added `requirePinMigration` to the auth store. `onRehydrateStorage` sets it when it force-clears a production `skipPin`; `setPin.tsx` uses it to route those DEFAULT_PIN users straight to the `set` stage (no `verify` prompt for a PIN they never chose). `handleSetPin` re-encrypts accounts from the old DEFAULT_PIN hash to the new PIN and clears the flag.

**5. Duress success loops back to lock screen** — The duress branch now `setLockTriggered(false)`, `setJustUnlocked(true)`, `resetPinTries()` before `router.push('/')`, so the `_layout.tsx` guard no longer redirects back to `/unlock`.

Added unit tests for `stripAllRpcCredentials` and `clearNostrFollowCaches` (and enabled immer transform in jest so immer-based stores are testable). `pnpm run check`, `typecheck`, and `test:unit` (1432 tests) all pass.

Psycarlo (approved):

PR #474, opened 2026-08-04
feat: strengthen PIN key derivation, add PIN lengths and passphrase meter

pedromvpg:
## Summary

Follow-up to the BIP39/PBKDF2 discussion (and audit finding: 4-digit PIN + PBKDF2 10k as the app's encryption root). The stored PIN digest doubles as the AES-256 key for every account seed in the app — with PBKDF2-SHA256 at 10k iterations over a fixed 4-digit PIN, an extracted keystore (root/backup) allowed brute-forcing every seed in seconds on a GPU. This PR attacks the problem on three axes:

**1. Memory-hard KDF with seamless migration** (`utils/pinKdf.ts`)
- New digests derive with **Argon2id** (RFC 9106 first-choice: m=64 MiB, t=3, p=4) when the native module is present, falling back to **scrypt** N=2^15/r=8/p=1 (~32 MiB), then **PBKDF2-SHA256 600k** (OWASP). Availability is probed at runtime, so any environment keeps working.
- The KDF config is stored next to each digest (`PIN_KDF_KEY`, duress has its own); verification always uses the *stored* config, so old digests keep verifying.
- **Migration**: on successful PIN auth (unlock or gated reveal), a legacy digest is re-derived under the current best KDF and all account key secrets are re-encrypted to it — two-phase (decrypt-all before any write; digest/config swap last), so an interrupted migration can't strand secrets between keys. Verified by unit tests.
- PIN/duress comparisons are now timing-safe.

**2. User-selectable PIN length (4–8 digits)** — the real entropy fix: 4 digits is 10⁴ candidates no matter the KDF. Length is chosen at set/change time, persisted, and the lock screen, duress flow, and PIN input adapt. Existing 4-digit users keep their length until they change it.

**3. BIP39 passphrase strength meter** — seed derivation is protocol-fixed at 2048 rounds (~11 bits of cost), so the passphrase is the *whole* defense once an attacker holds the mnemonic. The entry field now shows a live pessimistic strength estimate (pools capped hard for dictionary words, repeats, and trivial passwords) with a weak-passphrase warning.

Also fixed along the way: the `setDuressPin` store action regenerated the shared salt (would have silently invalidated the main PIN — audit finding #9's latent half), and the duress/main equality check now derives under the main PIN's own config.

**Explicitly out of scope** (noted for follow-up): Secure Enclave / Android Keystore key wrapping (native work), duress PIN KDF auto-migration (impossible without the plaintext — legacy duress digests keep verifying under their config until re-set), `DEFAULT_PIN`/`skipPin` handling (owned by #471/#472).

Note: touches some of the same files as #471/#472 (`crypto.ts` adjacent, `auth.ts`, `setPin.tsx`, `SSPinAuth.tsx`) — whichever lands second needs a small rebase, mostly import blocks.

## Test plan

- [x] `pnpm run check` — 0 errors
- [x] `pnpm --filter @satsigner/mobile typecheck` — clean
- [x] `pnpm --filter @satsigner/mobile test:unit` — 1461 pass (+35 new: pinKdf derive/serialize/migrate/fallbacks, auth store setPin/validatePin/legacy digest/duress salt, passphrase model, pin length helpers)
- [ ] Device: fresh install — set a 6-digit PIN, kill, unlock (migration is a no-op on new installs); wrong PIN fails
- [ ] Device (upgrade path): install previous build with a 4-digit PIN + accounts, upgrade, unlock once (migrates), kill, unlock again — accounts/seeds intact
- [ ] Device: set duress PIN (matches main length), enter it at lock — wipe occurs; main PIN change afterwards still verifies
- [ ] Device: change PIN from settings — verify stage shows the stored length, new length selectable, accounts still decrypt
- [ ] Device: import mnemonic with passphrase — strength hint responds (weak for "bitcoin"/single words, strong for long mixed)

---

**Update (2nd commit):** also folded in the entropy-hygiene fix from the dependency audit — `randomUuid()` (account ids, tx-builder local ids, NIP-46 session ids) used `react-native-uuid`'s `Math.random()` rng; it now draws CSPRNG bytes via `uuid.v4({ random })` (UUID format unchanged, +3 unit tests). With this, every randomness consumer in the app is CSPRNG-backed.


Psycarlo:
@pedromvpg ~Resolve conflicts please.

Psycarlo:
The core KDF work is solid (Argon2id → scrypt → PBKDF2 preference order, per-digest config persistence, re-encrypt-before-commit ordering inside `migratePinKdfIfNeeded`). The problems are all in integration. All findings below were verified against both the PR branch and master.

## Critical — permanent data loss

**1. PIN change commits the new salt/digest before re-encryption, with no rollback** — `app/setPin.tsx:174`
`setPin()` overwrites `SALT_KEY`/`PIN_KEY`/`PIN_KDF_KEY` first; `reEncryptAccounts` runs after, with no try/catch. If the app is killed or re-encryption throws mid-loop, the old salt is destroyed and every secret is permanently undecryptable. Master did this in two phases (`derivePinMaterial` → re-encrypt inside try/catch with `auth.pinChangeError` abort → `commitPinMaterial`); this PR deletes that ordering. Also: the digest comparison at lines 176–180 is dead code — `setPin` always draws a fresh salt, so re-entering the same PIN still re-encrypts and re-opens this window.

**2. KDF migration strands nostr and LND/RPC secrets** — `hooks/useKdfMigration.ts:16`
`reEncryptAllSecrets` only re-encrypts account key secrets, but `useReEncryptAccounts` (the PIN-change path) also calls `reEncryptNostrSecrets` and `reEncryptServiceSecrets`. After migration commits the new digest, the old PBKDF2-10k digest is gone forever — nostr nsecs and LND/RPC credentials become permanently undecryptable on first unlock after upgrade.

## Critical — duress PIN broken four ways

**3. Legacy duress PINs never verify after upgrade** — `components/SSPinAuth.tsx:89`
Pre-PR duress digests were derived with `SALT_KEY_DURESS`; verification now derives with the main `SALT_KEY`. No migration exists (`migratePinKdfIfNeeded` intentionally skips duress), and `SALT_KEY_DURESS` is no longer read anywhere. Under coercion the duress PIN does not wipe — it counts as a failed main-PIN attempt.

**4. Setting a duress PIN never enables it** — `app/setDuressPin.tsx:85`
The screen writes `DURESS_PIN_KEY`/`DURESS_KDF_KEY` directly but never sets `duressPinEnabled` (master called the store's `setDuressPin`, which does `set({ duressPinEnabled: true })`; that action now has zero production callers). `SSPinAuth` gates the duress branch on the flag, so a duress PIN set from onboarding is inert.

**5. Changing the main PIN silently invalidates the duress PIN** — `store/auth.ts:165`
`setPin` regenerates the shared salt; the duress digest (derived under the old salt) is never re-derived or cleared, and `duressPinEnabled` stays true. The comment in `setDuressPin` even acknowledges this ("vice versa on the next setPin") but nothing handles it. On master this couldn't happen — duress had its own salt.

**6. Duress wipe leaves secrets in SecureStore** — `utils/secureWipe.ts:25`
The deleted `useSecureWipe` hook called `deleteAllNostrSecretsForWipe`, `deleteLndSecretsSafe`, `deleteAllRpcCredentialsSafe`, and deleted `SALT_KEY_DURESS`. The rewrite drops all four — it only clears in-memory/zustand state for those. After a coerced wipe, nostr nsecs, LND secrets, and node RPC credentials remain extractable from SecureStore.

## High — lockout bugs

**7. Wrong verify attempt shrinks the PIN input to 4 cells** — `app/setPin.tsx:128`
The verify stage loads the stored length (4–8), but `handleVerifyPin` resets with `emptyPin()` (default 4). A user with a 6-digit PIN who mistypes once gets a 4-cell input that can never match — PIN change blocked until remount. Same bug on the Clear button (line 358).

**8. The scrypt fallback always throws at real parameters** — `utils/pinKdf.ts:75`
quick-crypto forwards its default `maxmem` (32 MiB) to `EVP_PBE_scrypt`, which needs `128·r·(N+2) + 128·r·p` = 33,557,504 bytes for N=32768/r=8 — over the limit (the JS-side check passes at exactly the boundary; verified in `HybridScrypt.cpp`). The availability probe uses N=2, so it succeeds and scrypt gets selected wherever argon2 is unavailable — then every real derivation throws and the user can never set or verify a PIN. The unit test masks this by passing `maxmem: 256 MiB` (`pinKdf.test.ts:114`). Pass `maxmem` explicitly or probe with real parameters.

## Medium — performance

**9. KDF runs synchronously on the JS thread** — `utils/pinKdf.ts:216`
`derivePinDigest` wraps `argon2Sync` (64 MiB, t=3, p=4) in `Promise.resolve` — roughly 0.5–2s of frozen UI per unlock on mid-range devices, doubled when a duress PIN is enabled (duress + main digests both derived even when configs and salt are identical). quick-crypto has async variants that run on the native thread pool.

## Minor

**10. Conventions** — hardcoded English toast `'Normal PIN must be set before setting Duress PIN'` (`app/setDuressPin.tsx:71`, CLAUDE.md rule 9); inline IIFE with nested ternaries (`components/SSSeedWordsInput.tsx:493`, rules 15/8); `as PinKdfConfig` casts in `utils/pinKdf.ts` default branches (rule 6 — a `never` exhaustiveness check works without casting).

Psycarlo (approved):

PR #481, opened 2026-08-08
feat: Nostr DMs (NIP-04/17), security report channel, diagnostics, signing crash fixes

pedromvpg:
Stacked on #474 (merge that first; this retargets to master automatically after).

Implements the first phases of `docs/plans/nostr-dms-and-relays.md` (included in this branch) and fixes several crashes found while testing on a physical device.

## Nostr DMs (NIP-04 + NIP-17)

The identity chat tabs were 'coming soon' placeholders; they now work:

- **Conversation lists** per protocol tab (peer, preview, unread badge, time) and a **thread screen** with inverted bubble list + composer (`components/chat/SSChatThread`, built to be shared with the devices group chat later)
- **New message** modal with contacts + npub paste
- NIP-17 via the existing gift-wrap primitives; NIP-04 via new `createKind4`/`subscribeToKind4` on NostrAPI (both sign inline — no signer needed at publish)
- New `nostr_chat_messages` table (SCHEMA_V9), dedup by (id, identity) so relay redelivery can't double-render; focus-scoped subscriptions
- Content is local plaintext for v1; the plan doc's envelope-encrypted store lands with #471. Requests bucket / outbox routing / relay supervisor are documented follow-ups

## Security report channel

- **Settings → About → Report a security issue**: NIP-17 gift wrap to the project npub (the SECURITY.md channel from #477), **anonymous by default** (throwaway key, no reply channel) or from the active identity (report lands in the chat thread so we can reply)
- Shows the publish relays up front, dismisses keyboard on send, and confirms with the exact sent message for validation

## Developer diagnostics (Settings → Developer → Diagnostics)

One-click checks, each also covered by unit/integration tests so they run in CI without the app: crypto roundtrip (AES/PBKDF2), **entropy collision test against the device's live RNG** (50k UUIDs + 20k IVs — Trust-Wallet-2023-class detector), PIN KDF, secure store, SQLite integrity, NIP-17 wrap/unwrap, and a **live relay roundtrip** that publishes a gift wrap to the project security npub and verifies retrieval + decryption (falls back to well-known relays when none configured).

## Crash fixes (all reproduce on master)

- **Signing: 'undefined is not a function'** — two compounding bugs:
  1. `psbt.txid()` was called unguarded though the store's psbt is not always a BDK Psbt (mocks/foreign implementations) → txid now derives from the PSBT bytes (`extractTransactionIdFromPSBT`), guarded fallback kept. A __DEV__ probe logs the rogue object's shape if it recurs
  2. **`TypedArray.prototype.toReversed()` doesn't exist in Hermes** — `txInput.hash.toReversed()` in PSBT extraction was the actual sign blocker. Replaced all remaining bare TypedArray call sites (psbt.ts ×2, electrum.ts, energy.tsx ×4) with the codebase's established `Buffer.from(x).reverse()` pattern
- **First-ever PIN set hung after confirm** (fresh installs): hardened `getPin()` throws when no PIN exists; the re-encrypt prefetch now tolerates null. Introduced by the #472 hardening — affects production fresh installs on master
- **Publish raced relay handshakes**: NDKRelay.publish rejects queued publishes after ~2.5s, so slow mobile TLS failed every publish ('failed to publish on any relay'). publishEvent now waits for ≥1 connected relay and publishes only to connected sockets
- **kind0 profiles never loaded** for fresh identities: `relayConnected` defaulted unset (fetches gate on it) and relay resolution fell through to empty lists — identities now default connected, and resolution falls back to indexing relays

## DevEx

- **Dev error reporting**: crashes + unhandled rejections now log message + raw stack immediately, then Metro-symbolicated source frames (file:line:column) in the console — the same modules LogBox uses. No more 'TypeError with no location'
- jest: `nostr-tools` added to transform list; new unit tests use the real module (the shared mock stubs exactly what these paths test)
- pnpm patch: `uniffi-bindgen-react-native@0.31.0-3` now exports `./package.json` (its generated CMakeLists resolves headers via `require.resolve('.../package.json')`, which the exports map blocked → 'UniffiCallInvoker.h not found')
- Added missing `@ubjs/core` runtime dep (imported by react-native-payjoin's generated code, never declared)

- **Receive QR zero-amount edge**: `amountSats=0` emitted `amount=0`, which some wallets reject; zero now omits the amount param

## Receive QR validity

New `receiveQrUriValidity.test.ts` round-trips `buildReceiveQrUri` output through the **real** `bip-321` parser (the shared jest mock only checks superficially): plain address, exact sats amounts, special-character labels, 1-sat precision, and payjoin session URI rewrites all parse valid. Mixed-case is preserved on purpose — percent-encoded params (and BIP-77 `pj=`) are case-sensitive.

## Test plan

- [x] 165 suites / 1795 unit tests pass; 41 integration tests pass (incl. live NIP-17 relay roundtrip against public relays)
- [x] tsc clean; expo export (android) bundles
- [x] On-device (Pixel 9): fresh-install PIN set, sign pasted PSBT, nostr profile load, diagnostics checks, security report send
- [ ] Maestro flows for chat (follow-up)


Psycarlo:
# PR #481: findings to fix before merge

## Blocking

### 1. NIP-17 messages corrupted for non-ASCII text

`apps/mobile/api/nostr.ts:1298`

`createKind1059` pre-encodes content with `unescape(encodeURIComponent(content))`, but nothing in the receive path inverts it. `unwrapNip59EventOrNull` returns the rumor verbatim and `utils/nostrChat.ts:399` stores it as-is. Any accented characters, CJK, or emoji arrive as mojibake, both for recipients and for self copies. Tests only cover ASCII.

Fix: drop the pre-encode (nostr-tools handles UTF-8 strings) or decode symmetrically on ingest.

### 2. Chat pipeline refcount leak

`hooks/useNostrChat.ts:48-60`, `utils/nostrChat.ts:330`

If the screen unmounts while `acquireChatPipeline` is still connecting, cleanup runs with `held === false` and never releases. The late `.then` still increments `refs`, so the orphaned pipeline keeps its subscriptions open forever. Also, `refs += 1` happens after awaits, so an identity switch mid-acquire credits the wrong pipeline.

### 3. Partial publish failures are swallowed

`api/nostr.ts:1758-1764`

`publishEvent` publishes only to relays connected within the wait window and throws only if all fail. A gift wrap can land on 1 of N intended inbox relays with no log or retry, silently losing DM delivery on flaky networks.

Fix: at minimum log failed and skipped relay URLs; ideally retry unconnected relays once connected.

### 4. Unread badges go stale after opening a thread

`hooks/useNostrChat.ts:208-213`, `db/mutations/nostrChat.ts:41`

`markChatThreadRead` updates the DB without emitting a chat listener event, and `useNostrChatConversations` never reloads on focus. Returning from a thread still shows the unread count until the next message arrives.

### 5. Unconditional console.log in production code

`utils/nostrChat.ts:34-37`

`chatLog` runs in production builds and logs npub prefixes, relay URLs, and message ids on every ingest. Violates the repo's no-console rule.

Fix: gate behind `__DEV__`.

Psycarlo (approved):

Issue #480, opened 2026-08-08
share QR codes via android/ios share menu


v4v2:
add button to share qr codes everywhere they appear in the app, via the native share menu

PR #487, opened 2026-08-25
Add share button for native share menu

v4v2:
Add SSShareButton below SSQRCode for address, invoice, descriptor, pubkey, PSBT, token, etc, so it can be shared via the native share menu.

- [x] Add reusable SSShareButton component
- [x] Use SShareButton for non-sensitive QR code content
- [x] Manually check for intended behavior accross affected pages


Psycarlo:
## Bug: sharing an animated QR only captures one frame

When a PSBT or ecash token is too big for a single QR, the screen cycles through chunks. The share button screenshots whatever chunk is on screen at that moment. The result is one fragment that can't be decoded on its own. The user gets no warning.

Happens in 4 places:

- `apps/mobile/components/SSPsbtTransport.tsx` (line ~216)
- `apps/mobile/app/.../signAndSend/previewTransaction.tsx` (line ~2652), affects RAW, UR and BBQR modes
- `apps/mobile/app/.../ecash/send.tsx` (line ~365)
- `apps/mobile/app/.../ecash/account/[id]/proof/[proofIndex]/index.tsx` (line ~533)

Possible fixes:

- Hide the share button when the QR is multi-part / animated, or
- Share the full data as text instead of a screenshot, or
- Pause the animation and share all frames

## Cleanup: duplicated boilerplate

The same `<View collapsable={false} ref={qrRef}><SSQRCode /></View>` + `<SSShareButton qrRef={qrRef} />` block is copy-pasted in 22 screens. Extract a single wrapper component (e.g. `SSShareableQR`) so the bug above can be fixed in one place.

Psycarlo (approved):

v4v2:
closes #480

PR #501, opened 2026-09-16
fix: use distinct board and offboard icons in Ark history

pedromvpg:
## Summary
- Treat Ark board and offboard as their own movement kinds so they no longer reuse send/receive icons.
- Use the wallet-menu board/offboard icons (green/red) on history, movement detail, and the VTXO selection bar.
- Drop the minus on outgoing amounts and baseline-align the sats label with the amount.

## Test plan
- [ ] Open an Ark wallet and confirm board/offboard in history use the circle icons, not send/receive arrows
- [ ] Open a board movement and an offboard movement and confirm the same icons and kind labels
- [ ] Confirm send/receive still use the existing incoming/outgoing icons
- [ ] Confirm the settings menu still uses the gray board/offboard icons


Psycarlo (approved):
LGTM

